PR #17, opened 2024-12-03
Added match function to the backend

ghost:
# Context

Top of the [features shopping list](https://uktrade.atlassian.net/wiki/spaces/CDL/pages/4409622534/Matchbox+feature+shopping+list) and vital for replicating the existing Company Matching endpoint using Matchbox is `match()`.

Given an ID in a source dataset, it should return the paired keys from the target dataset from the perspective of a point of truth.

While there's plenty of edge cases and ways we might extend this, this is the vanilla implementation so we have something to work with.

* Adds `match()` as a function throughout the library: client, server base class, PostgreSQL adapter
* Adds simple unit tests based on identifying "Revolution Inc" across the warehouse data
* Adds indices to `Clusters` and `Contains` to facilitate speed on this query
* Modifies `batch_ingest()` to permit use of indices via a new `isolate_table` function
* Adds some helper function to visualise the data subgraph -- I used them to help me debug and felt they were useful enough to leave in
* Minor change to `Models.get_lineage_to_dataset()`: the dataset model is now also returned, with accompanying change in `query()` logic

## Guidance to review

Focus on `matchbox.server.postgresql.utils.query` -- this is where the biggest, most confusing changes are.

* My unit tests lack some coverage. I don't have any examples of simple `1:1` because my source data has no examples of this. I think this is okay for now -- do you agree?
    * Note `1:0` is now covered
* I chose to implement the IDs as sets rather than `str | list[str] | None` because I think it's more elegant
* "Source" and "target" aren't language used elsewhere in the system, but I think it's justified as this is a reasonably unique operation for the system. Do you agree with this language?
* I considered implementing the API out of scope. Do you agree?

## Relevant links

* [Feature shopping list](https://uktrade.atlassian.net/wiki/spaces/CDL/pages/4409622534/Matchbox+feature+shopping+list)
* [Matchbox in the CDL](https://uktrade.atlassian.net/wiki/spaces/CDL/pages/4405231689/Matchbox+in+the+CDL), where this function was ideated

## Checklist:

### DIFF
--- a/test/fixtures/data.py
+++ b/test/fixtures/data.py
@@ -135,13 +135,23 @@ def revolution_inc(
         duns_companies["company_name"].str.contains("Revolution", case=False)
     ]["id"].tolist()
 
-    cdms_ids = cdms_companies[cdms_companies["crn"] == crn_ids[0]]["id"].tolist()
+    revolution_crn = crn_companies[
+        crn_companies["company_name"].str.contains("Revolution", case=False)
+    ]["crn"].iloc[0]
 
-    return {
+    cdms_ids = cdms_companies[cdms_companies["crn"] == revolution_crn]["id"].tolist()
+
+    revolution = {
         "crn": [str(id) for id in crn_ids],
         "duns": [str(id) for id in duns_ids],
         "cdms": [str(id) for id in cdms_ids],
     }
+
+    assert len(revolution.get("crn", [])) == 3
+    assert len(revolution.get("duns", [])) == 1
+    assert len(revolution.get("cdms", [])) == 2
+
+    return revolution
 
 
 @pytest.fixture(scope="session")
@@ -167,13 +177,23 @@ def winner_inc(
         duns_companies["company_name"].str.contains("Winner", case=False)
     ]["id"].tolist()
 
-    cdms_ids = cdms_companies[cdms_companies["crn"] == crn_ids[0]]["id"].tolist()
+    winner_crn = crn_companies[
+        crn_companies["company_name"].str.contains("Revolution", case=False)
+    ]["crn"].iloc[0]
 
-    return {
+    cdms_ids = cdms_companies[cdms_companies["crn"] == winner_crn]["id"].tolist()
+
+    winner = {
         "crn": [str(id) for id in crn_ids],
         "duns": [str(id) for id in duns_ids],
         "cdms": [str(id) for id in cdms_ids],
     }
+
+    assert len(winner.get("crn", [])) == 3
+    assert len(winner.get("duns", [])) == 0
+    assert len(winner.get("cdms", [])) == 2
+
+    return winner
 
 
 @pytest.fixture(scope="function")


src/matchbox/common/db.py
```diff
@@ -33,6 +33,28 @@
...
+    @model_validator(mode="after")
+    def found_or_none(self) -> "Match":
+        if self.cluster is None and (self.source_id or self.target_id):
+            raise ValueError(
+                "A match must have a cluster if source_id or target_id is set."
+            )
+        elif self.cluster is not None and not (self.source_id or self.target_id):
+            raise ValueError(
+                "A match must have source_id or target_id if cluster is set."
+            )
```
lmazz1-dbt, line 54: not both?

lmazz1-dbt (commented):
I agree that the API endpoint is out of scope, but I'd like to see more tests

src/matchbox/server/base.py
```diff
@@ -251,6 +251,16 @@ def query(
         limit: int = None,
     ) -> PandasDataFrame | ArrowTable | PolarsDataFrame: ...
 
+    @abstractmethod
+    def match(
+        self,
+        source_id: str,
+        source: str,
+        target: str | list[str],
+        model: str,
```
lmazz1-dbt, line 260: One day this should be optional but I guess we're not there yet

src/matchbox/server/postgresql/utils/db.py
```diff
@@ -50,6 +57,44 @@ def get_model_subgraph(engine: Engine) -> rx.PyDiGraph:
...
+    """Retrieves the complete data subgraph as a PyDiGraph."""
+    G = rx.PyDiGraph()
+    nodes = {}
+
+    with Session(engine) as session:
+        sources = {source.model: source for source in session.query(Sources).all()}
+
+        for source in sources.values():
+            source_id = f"{source.schema}.{source.table}"
+            if source_id not in nodes:
```
lmazz1-dbt, line 70: are we expecting duplicates?

src/matchbox/server/postgresql/utils/db.py
```diff
@@ -50,6 +57,44 @@ def get_model_subgraph(engine: Engine) -> rx.PyDiGraph:
     return G
 
 
+def get_data_subgraph(engine: Engine) -> rx.PyDiGraph:
```
lmazz1-dbt, line 60: Is this worth testing? By setting up the DB at the index level, and then checking against a manually constructed expectation object. Though every one of these tests that tear down and set up the DB takes some time to run

src/matchbox/server/postgresql/utils/db.py
```diff
@@ -50,6 +57,44 @@ def get_model_subgraph(engine: Engine) -> rx.PyDiGraph:
...
+
+        for cluster in session.query(Clusters).all():
+            cluster_id = cluster.hash
+            if cluster_id not in nodes:
+                cluster_idx = G.add_node({"id": cluster_id, "type": "cluster"})
+                nodes[cluster_id] = cluster_idx
+
+            if cluster.id is not None and cluster.dataset is not None:
+                source = sources.get(cluster.dataset)
+                if source:
```
lmazz1-dbt, line 82: can it be the case that `cluster.dataset is not None` and `source is None` ?

src/matchbox/server/postgresql/utils/db.py
```diff
@@ -50,6 +57,44 @@ def get_model_subgraph(engine: Engine) -> rx.PyDiGraph:
...
+                    data_id = str(cluster.id)
+                    data_idx = G.add_node({"id": data_id, "type": "data"})
```
lmazz1-dbt, line 84: Just curious: why do you have a separate node for a data point and a cluster of 1?

test/fixtures/data.py
```diff
@@ -106,12 +106,42 @@ def cdms_companies(all_companies: DataFrame) -> DataFrame:
...
+        crn_companies["company_name"].str.contains("Revolution", case=False)
+    ]["id"].tolist()
+
+    duns_ids = duns_companies[
+        duns_companies["company_name"].str.contains("Revolution", case=False)
+    ]["id"].tolist()
+
+    revolution_crn = crn_companies[
+        crn_companies["company_name"].str.contains("Revolution", case=False)
+    ]["crn"].iloc[0]
```
lmazz1-dbt, line 134: can you re-use the query on crn_companies a few lines above?

test/fixtures/data.py
```diff
@@ -106,12 +106,42 @@ def cdms_companies(all_companies: DataFrame) -> DataFrame:
...
+
+    UUIDs are converted to strings to mirror how Matchbox stores them.
+    """
+    crn_ids = crn_companies[
+        crn_companies["company_name"].str.contains("Revolution", case=False)
+    ]["id"].tolist()
+
+    duns_ids = duns_companies[
+        duns_companies["company_name"].str.contains("Revolution", case=False)
+    ]["id"].tolist()
```
lmazz1-dbt, line 130: the tests later (I think) asssume that there are multiple crn IDs and one duns ID. Can we call this out here? I also don't like that this is so tied to a large CSV shared by all tests. Might have to be part of later work on improving tests though

test/server/test_adapter.py
```diff
@@ -541,6 +541,72 @@ def test_query_with_link_model(self):
...
+        )
+
+        assert isinstance(res, Match)
+        assert res.source == str(duns_wh)
+        assert res.target == str(crn_wh)
+        assert res.source_id == set(revolution_inc["duns"])
+        assert res.target_id == set(revolution_inc["crn"])
+
+    def test_match_many_to_one(self, revolution_inc: dict[str, list[str]]):
+        """Test that matching data works when the source has more possible IDs."""
```
lmazz1-dbt, line 567: This test looks a bit obscure to me. On line 585 we're asserting that source_id has multiple elements, right?

src/matchbox/server/postgresql/utils/db.py
lmazz1-dbt: Untested code in the codebase in main is tech debt. I don't mind if it goes on a branch or in a ticket or on Confluence

src/matchbox/server/postgresql/utils/query.py
```diff
@@ -364,3 +401,271 @@ def query(
...
+        # Get source, target and truth resolutions
+        source_resolution = source_to_dataset_resolution(source, session)
+        target_resolutions = []
+        for target in targets:
+            target_resolutions.append(source_to_dataset_resolution(target, session))
+        truth_resolution = (
+            session.query(Resolutions).filter(Resolutions.name == resolution).first()
+        )
+        if truth_resolution is None:
+            raise MatchboxResolutionError(f"Resolution {resolution} not found")
```
lmazz1-dbt, line 593: that's the same as the default error message, just pass resolution as an argument

src/matchbox/server/postgresql/utils/query.py
```diff
@@ -364,3 +401,271 @@ def query(
...
+            source_ids = {
+                id
+                for _, dataset_hash, id in matches
+                if dataset_hash == source_resolution.hash
+            }
+            target_ids = {
+                id
+                for _, dataset_hash, id in matches
+                if dataset_hash == target_resolution.hash
+            }
```
lmazz1-dbt, line 660: could this not be part of the SQL query?

src/matchbox/server/postgresql/utils/query.py
```diff
@@ -364,3 +401,271 @@ def query(
...
+        result = []
+        for target_resolution in target_resolutions:
+            # Get source/target table names
+            target_schema, target_table = next(
+                (schema, table)
+                for schema, table in target_pairs
+                if session.get(Sources, target_resolution.hash).schema == schema
+                and session.get(Sources, target_resolution.hash).table == table
+            )
+            target_name = f"{target_schema}.{target_table}"
```
lmazz1-dbt, line 648: Is this equivalent to a list comprehension followed by `[0]`?

test/fixtures/data.py
```diff
@@ -144,6 +144,38 @@ def revolution_inc(
...
+    Winner Inc. as it exists across all three datasets.
+
+    UUIDs are converted to strings to mirror how Matchbox stores them.
+
+    Based on the above fixtures, should return:
+
+    * Three CRNs
+    * Zero DUNS
+    * Two CDMS
+    """
```
lmazz1-dbt, line 161: could you... assert this is true? Not sure how I feel abbout assertions in a fixture, but it would increase my confidence in the stability of tests without having to just trust that the docstring is true now and in the future

lmazz1-dbt (approved):

PR #22, opened 2024-12-05
Add the ability to choose the columns we index on

ghost:
# Context

Matchbox currently indexes every column. This makes it difficult to control how data is hashed, and is vulnerable to column renames or reorderings. [Ticket](https://uktrade.atlassian.net/jira/software/projects/MTCHBX/boards/1123?selectedIssue=MTCHBX-58). [Feature shopping list](https://uktrade.atlassian.net/wiki/spaces/CDL/pages/4409622534/Matchbox+feature+shopping+list).

* Adds column specification to `datasets.toml` config file. You can configure:
    * Order of columns
    * Alias of columns
    * Datatype of columns
    * A special `*` character as a shorthand for "select all"
* Adds table aliasing, column specification and hashing to `Source` object, ensuring private data isn't sent to the server
* Adds `Sources.alias` and `Sources.indices` to PostgreSQL backend, enabling reconstruction of `Source` by client if they have permission to access the table
* Adds a check and warning to `query` if a non-indexed column is selected
* Adds unit tests to cover these additions

## Guidance to review

* Check unit tests pass
* Confirm you're happen with the two extra columns in `Sources`, especially the datatype of the column indexes (`JSONB` with base64-encoded hashes, as pure bytes isn't permitted)
* Confirm you're happy with the security model of what's stored vs what's retrieved
    * If you don't have warehouse creds, you can't know what columns have been used to index in Matchbox
    * BUT, internal users who have access and know the table can recover them
    * AND hypothetical external uses can check whether their hashes are matching ours as they tinker with federation settings
* Note `SourceColumn.type` isn't actually used anywhere yet -- the hashing algorithm casts everything to string anyway

## Checklist:

### DIFF
--- a/src/matchbox/common/db.py
+++ b/src/matchbox/common/db.py
@@ -237,7 +237,7 @@ class Source(BaseModel):
         Handles three scenarios:
             1. No columns specified - all columns except primary key are indexed
             2. Indices from database - uses existing column hash information
-            3. Columns specified in TOML - uses specified columns with optional '*'
+            3. Columns specified in TOML - specified columns are indexed
         """
         # Initialise warehouse and get table metadata
         warehouse = (
@@ -280,14 +280,16 @@ class Source(BaseModel):
 
         # Case 3: Columns from TOML
         local_columns = []
-        star_index = None
 
         # Process TOML column specifications
-        for i, column in enumerate(index_data):
-            if column["literal"] == "*":
-                star_index = i
-                continue
-            local_columns.append(SourceColumn(**column, indexed=True))
+        for column in index_data:
+            local_columns.append(
+                SourceColumn(
+                    literal=column["literal"],
+                    alias=column.get("alias", column["literal"]),
+                    indexed=True,
+                )
+            )
 
         # Match remote columns with local specifications
         indexed_columns = []
@@ -296,23 +298,14 @@ class Source(BaseModel):
         for remote_col in remote_columns:
             matched = False
             for local_col in local_columns:
-                if remote_col == local_col:
-                    if local_col.type is None:
-                        local_col.type = remote_col.type
+                if remote_col.literal == local_col.literal:
                     indexed_columns.append(local_col)
                     matched = True
                     break
             if not matched:
                 non_indexed_columns.append(remote_col)
 
-        # Handle wildcard insertion
-        if star_index is not None:
-            for col in non_indexed_columns:
-                col.indexed = True
-            indexed_columns[star_index:star_index] = non_indexed_columns
-            data["db_columns"] = indexed_columns
-        else:
-            data["db_columns"] = indexed_columns + non_indexed_columns
+        data["db_columns"] = indexed_columns + non_indexed_columns
 
         return data
 
--- a/test/client/test_admin.py
+++ b/test/client/test_admin.py
@@ -40,17 +40,9 @@ def source_toml(source: Source, index: list[dict[str, str]]) -> str:
             {"literal": "company_name"},
             {"literal": "crn"},
         ],
-        [{"literal": "company_name", "type": "VARCHAR", "alias": "name"}],
-        [
-            {"literal": "company_name"},
-            {"literal": "*"},
-        ],
-        [
-            {"literal": "*"},
-            {"literal": "company_name"},
-        ],
+        [{"literal": "company_name", "alias": "name"}],
     ),
-    ids=["vanilla", "alias_and_type", "star_end", "star_start"],
+    ids=["vanilla", "alias"],
 )
 def test_load_datasets_from_config(
     index: list[dict[str, str]],
@@ -76,9 +68,7 @@ def test_load_datasets_from_config(
 
     # Helper variables
     source = config.get(crn.alias)
-    named = [idx["literal"] for idx in index if idx["literal"] != "*"]
-    has_star = any(idx["literal"] == "*" for idx in index)
-    star_pos = next((i for i, idx in enumerate(index) if idx["literal"] == "*"), None)
+    named = [idx["literal"] for idx in index]
     col_names = [col.literal.name for col in source.db_columns]
 
     # Test 1: Core attributes match
@@ -91,33 +81,16 @@ def test_load_datasets_from_config(
     # Test 2: All non-pk columns present
     assert set(col_names) == {"company_name", "crn", "id"} - {source.db_pk}
 
-    # Test 3: Column indexing
-    for col in source.db_columns:
-        assert col.indexed == (has_star or col.literal.name in named)
-
-    # Test 4: Aliases and types match
+    # Test 3: Aliases match
     for idx in index:
-        if idx["literal"] == "*":
-            continue
         col = next(c for c in source.db_columns if c.literal.name == idx["literal"])
         assert col.alias.name == idx.get("alias", idx["literal"])
-        assert col.type == idx.get("type", col.type)
 
-    # Test 5: Column ordering
-    if star_pos is None:
-        for i, name in enumerate(named):
-            assert col_names[i] == name
-    else:
-        for i, idx in enumerate(index):
-            if idx["literal"] != "*":
-                if i < star_pos:
-                    assert col_names[i] == idx["literal"]
-                else:
-                    star_col_count = len(col_names) - len(index) + 1
-                    assert col_names[i + star_col_count - 1] == idx["literal"]
+    # Test 4: Column ordering
+    for i, name in enumerate(named):
+        assert col_names[i] == name
 
-    # Test 6: column equalities
-
+    # Test 5: column equalities
     assert source.db_columns[0] != source.db_columns[1]
     assert source.db_columns[0] == source.db_columns[0]
     assert source.db_columns[1].literal.hash == source.db_columns[1]


src/matchbox/common/db.py
```diff
@@ -83,6 +84,25 @@ def engine(self) -> Engine:
             self.test_connection()
         return self._engine
 
+    def __str__(self):
```
lmazz1-dbt, line 87: note to self: why `__str__` and not `__repr__`?

src/matchbox/common/db.py
```diff
@@ -116,17 +130,103 @@ def from_engine(cls, engine: Engine, alias: str | None = None) -> "SourceWarehou
...
+
+    @property
+    def hash(self) -> bytes:
+        """Generate a unique hash based on the column name."""
+        return HASH_FUNC(self.name.encode("utf-8")).digest()
+
+    @property
+    def base64(self) -> str:
+        """Generate a base64 encoded hash based on the column name."""
+        return base64.b64encode(self.hash).decode("utf-8")
```
lmazz1-dbt, line 146: Doing the same in my PR, we might want to standardise this later

src/matchbox/common/db.py
```diff
@@ -116,17 +130,103 @@ def from_engine(cls, engine: Engine, alias: str | None = None) -> "SourceWarehou
...
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    literal: SourceColumnName = Field(
+        description="The literal name of the column in the database."
+    )
+    alias: SourceColumnName = Field(
+        default_factory=lambda data: SourceColumnName(name=data["literal"].name),
+        description="The alias to use when hashing the dataset in Matchbox.",
+    )
+    type: str | None = Field(
```
lmazz1-dbt, line 161: Note to self: what happens if this isn't specified?

src/matchbox/common/db.py
```diff
@@ -116,17 +130,103 @@ def from_engine(cls, engine: Engine, alias: str | None = None) -> "SourceWarehou
...
+        default=None, description="The type to cast the column to before hashing data."
+    )
+    indexed: bool = Field(description="Whether the column is indexed in the database.")
```
lmazz1-dbt, line 164: Note to self: review how this is used

src/matchbox/common/db.py
```diff
@@ -116,17 +130,103 @@ def from_engine(cls, engine: Engine, alias: str | None = None) -> "SourceWarehou
...
+
+    def __eq__(self, other: object) -> bool:
+        """Compare SourceColumn with another SourceColumn or bytes object.
+
+        Two SourceColumns are equal if:
+
+        * Their literal names match, or
```
lmazz1-dbt, line 171: What happens if two columns have the same alias but they're not the same column?

src/matchbox/common/db.py
```diff
@@ -116,17 +130,103 @@ def from_engine(cls, engine: Engine, alias: str | None = None) -> "SourceWarehou
...
+
+            return bool(self_hashes & other_hashes)
+
+        if isinstance(other, bytes):
+            return other in {self.literal.hash, self.alias.hash}
+
+        return NotImplemented
+
+    @field_validator("literal", "alias", mode="before")
+    def string_to_name(cls: "SourceColumn", value: str) -> SourceColumnName:
```
lmazz1-dbt, line 201: See my comment above about whether we need a separate `SourceColumnName`, feels like over-engineering

src/matchbox/common/db.py
```diff
@@ -116,17 +130,103 @@ def from_engine(cls, engine: Engine, alias: str | None = None) -> "SourceWarehou
...
+        if isinstance(value, str):
+            return SourceColumnName(name=value)
+        else:
+            raise ValueError("Column name must be a string.")
+
+
+class SourceIndex(BaseModel):
+    """The hashes of column names in the Matchbox database."""
```
lmazz1-dbt, line 209: Is this docstring correct? Can't see no hashes here

src/matchbox/common/db.py
```diff
@@ -116,17 +130,103 @@ def from_engine(cls, engine: Engine, alias: str | None = None) -> "SourceWarehou
...
+
+    literal: list[bytes]
+    alias: list[bytes]
```
lmazz1-dbt, line 212: Here we have an implicit coupling between two fields. Would something like this work?
```
class  SourceIndex(BaseModel):
  literal: bytes
  alias: bytes

source_indices: list[SourceIndex]
```

src/matchbox/server/postgresql/orm.py
```diff
@@ -152,9 +152,11 @@ class Sources(CountMixin, MBDB.MatchboxBase):
     model = Column(
         BYTEA, ForeignKey("models.hash", ondelete="CASCADE"), primary_key=True
     )
+    alias = Column(VARCHAR, nullable=False, unique=True)
     schema = Column(VARCHAR, nullable=False)
     table = Column(VARCHAR, nullable=False)
```
lmazz1-dbt, line 157: Would the combination of these also be unique?

src/matchbox/server/postgresql/adapter.py
```diff
@@ -354,10 +355,17 @@ def get_dataset(self, db_schema: str, db_table: str, engine: Engine) -> Source:
...
+                for index_type, index_b64_list in dataset.indices.items():
+                    dataset_indices[index_type] = [
+                        base64.b64decode(b64.encode("utf-8")) for b64 in index_b64_list
+                    ]
                 return Source(
+                    alias=dataset.alias,
                     db_schema=dataset.schema,
                     db_table=dataset.table,
                     db_pk=dataset.id,
+                    db_columns=dataset_indices,
```
lmazz1-dbt, line 368: To check my understanding: there is an assumption that the columns we use for matching are a subset of the columns indexed in the source warehouse? I think I find this 
```
db_columns = indices
```
semantically confusing

test/server/test_adapter.py
```diff
@@ -114,9 +114,24 @@ def test_get_dataset(self):
...
+        crn_retrieved = self.backend.get_dataset(
             db_schema=crn.db_schema, db_table=crn.db_table, engine=crn.database.engine
         )
+
+        assert crn.db_columns == crn_retrieved.db_columns
+
+        cols: dict[str, list[SourceColumn]] = {}
+
+        # Indexing isn't used in the custom equality check
+        for col in crn.db_columns + crn_retrieved.db_columns:
```
lmazz1-dbt, line 126: Why the plus operator, haven't we asserted they're the same a few lines above?

test/server/test_adapter.py
```diff
@@ -114,9 +114,24 @@ def test_get_dataset(self):
...
+            if col.literal.name not in cols:
+                cols[col.literal.name] = [col]
+            else:
+                cols[col.literal.name].append(col)
```
lmazz1-dbt, line 130: nitpicking, but this could be a defaultdict of type list without the if condition

test/server/test_adapter.py
```diff
@@ -114,9 +114,24 @@ def test_get_dataset(self):
...
+
+        for c1, c2 in cols.values():
+            assert c1.indexed == c2.indexed
+
```
lmazz1-dbt, line 134: Nitpicking, but you could use a default dict of type list instead of the if condition

src/matchbox/common/db.py
```diff
@@ -136,6 +236,94 @@ def __hash__(self) -> int:
...
+        Handles three scenarios:
+            1. No columns specified - all columns except primary key are indexed
+            2. Columns specified in TOML - uses specified columns with optional '*'
+            3. Indices from database - uses existing column hash information
+        """
+        # Initialise warehouse and get table metadata
+        warehouse = (
+            data["database"]
+            if isinstance(data["database"], SourceWarehouse)
+            else SourceWarehouse(**data["database"])
```
lmazz1-dbt, line 253: Why do you want to allow deferring the construction of a SourceWarehouse?

src/matchbox/common/db.py
```diff
@@ -116,17 +130,103 @@ def from_engine(cls, engine: Engine, alias: str | None = None) -> "SourceWarehou
...
+        """Generate a unique hash based on the column name."""
+        return HASH_FUNC(self.name.encode("utf-8")).digest()
+
+    @property
+    def base64(self) -> str:
+        """Generate a base64 encoded hash based on the column name."""
+        return base64.b64encode(self.hash).decode("utf-8")
+
+
+class SourceColumn(BaseModel):
```
lmazz1-dbt, line 149: Is a `SourceColumn` substantially different from a `SourceIndex`?

src/matchbox/common/db.py
```diff
@@ -136,6 +236,94 @@ def __hash__(self) -> int:
...
 
+    @model_validator(mode="before")
+    @classmethod
+    def hash_columns(cls, data: dict[str, Any]) -> "Source":
+        """Shapes indices data from either the backend or TOML.
+
+        Handles three scenarios:
+            1. No columns specified - all columns except primary key are indexed
+            2. Columns specified in TOML - uses specified columns with optional '*'
+            3. Indices from database - uses existing column hash information
```
lmazz1-dbt, line 247: small point: in the code these two are in the opposite order

src/matchbox/common/db.py
```diff
@@ -136,6 +236,94 @@ def __hash__(self) -> int:
...
+                    type=col.type,
+                    indexed=col in source_index.literal + source_index.alias,
+                )
+                for col in remote_columns
+            ]
+            return data
+
+        # Case 3: Columns from TOML
+        local_columns = []
+        star_index = None
```
lmazz1-dbt, line 291: The fact that we support both TOML as well as procedural definition of sources makes me once more confused about when we will and won't use the TOML file

src/matchbox/common/db.py
```diff
@@ -136,6 +236,94 @@ def __hash__(self) -> int:
...
+        # Match remote columns with local specifications
+        indexed_columns = []
+        non_indexed_columns = []
+
+        for remote_col in remote_columns:
+            matched = False
+            for local_col in local_columns:
+                if remote_col == local_col:
+                    if local_col.type is None:
+                        local_col.type = remote_col.type
```
lmazz1-dbt, line 309: why do we allow to specify the type at all if we're doing this?

src/matchbox/common/db.py
```diff
@@ -136,6 +236,94 @@ def __hash__(self) -> int:
...
+                non_indexed_columns.append(remote_col)
+
+        # Handle wildcard insertion
+        if star_index is not None:
+            for col in non_indexed_columns:
+                col.indexed = True
+            indexed_columns[star_index:star_index] = non_indexed_columns
+            data["db_columns"] = indexed_columns
+        else:
+            data["db_columns"] = indexed_columns + non_indexed_columns
```
lmazz1-dbt, line 323: My comment above applies here as well: is it worth being more conservative? How much are we gaining by having all these different options to index?

test/client/test_admin.py
```diff
@@ -0,0 +1,126 @@
...
+    with NamedTemporaryFile(suffix=".toml", delete=False) as temp_file:
+        temp_file.write(raw_toml.encode())
+        temp_file.flush()
+        temp_file_path = Path(temp_file.name)
+
+    # Ingest
+    config = load_datasets_from_config(temp_file_path)
+
+    # Helper variables
+    source = config.get(re.sub(r"[^a-zA-Z0-9]", "", crn.alias))
```
lmazz1-dbt, line 78: so aliases can't have underscores?

test/client/test_admin.py
```diff
@@ -0,0 +1,126 @@
...
+        if idx["literal"] == "*":
+            continue
+        col = next(c for c in source.db_columns if c.literal.name == idx["literal"])
+        assert col.alias.name == idx.get("alias", idx["literal"])
+        assert col.type == idx.get("type", col.type)
+
+    # Test 5: Column ordering
+    if star_pos is None:
+        for i, name in enumerate(named):
+            assert col_names[i] == name
```
lmazz1-dbt, line 109: I think this is redundant if you consider the logic below

test/client/test_admin.py
```diff
@@ -0,0 +1,126 @@
...
+    else:
+        for i, idx in enumerate(index):
+            if idx["literal"] != "*":
+                if i < star_pos:
+                    assert col_names[i] == idx["literal"]
+                else:
+                    star_col_count = len(col_names) - len(index) + 1
```
lmazz1-dbt, line 116: +1? Which you're then subtracting later?

src/matchbox/common/db.py
lmazz1-dbt: could you add a TODO?

src/matchbox/common/db.py
lmazz1-dbt: sure, but maybe you can  do
```python
from matchbox.common.hash import hash_to_str
@property
def base64(self) -> str:
    return hash_to_str(self.hash)
```

src/matchbox/common/db.py
lmazz1-dbt: What could happen is that the TOML could stay the same, but the table could change, and this would make the indexing of the hash of the table change. Stability of the hash as defined by the TOML feels like a feature, which the "star" breaks

src/matchbox/common/db.py
lmazz1-dbt: I think out

src/matchbox/common/db.py
lmazz1-dbt: Resolving as this is discussed in a different thread

test/client/test_admin.py
lmazz1-dbt: Still unclear. Why would there be a dot in aliases? Are we happy with, e.g. a curly brace?

lmazz1-dbt (approved):

PR #24, opened 2024-12-10
Feature/mtchbx 45

nimonika:
# Context

* Three endpoints for sources

## Guidance to review

The POST endpoint is based on the assumption that the client can send an entire file. This may not be a realistic assumption. For large files, the client will need to send chunks as multipart objects, which in turn will need to be sent as multiparts to s3. Unit tests have been written only for the two GET endpoints as the POST is more of an integration test

## Relevant links

* [List any useful links]

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] New and existing unit tests pass locally with my changes


src/matchbox/server/api.py
```diff
@@ -45,6 +46,21 @@ class CountResult(BaseModel):
...
+    schema: str
+    table: str
+    id: str
+    resolution: str | None = None
+
+
+class Sources(BaseModel):
+    """Response model for sources"""
+
+    sources: list[SourceItem]
```
lmazz1-dbt, line 61: should we start moving this stuff to common?

lmazz1-dbt, line 61: do we need a separate pydantic object? Can't we just return list[Sources]?

src/matchbox/server/api.py
```diff
@@ -76,18 +92,49 @@ async def clear_backend():
...
-
-@app.post("/sources/{hash}")
-async def add_source(hash: str):
-    raise HTTPException(status_code=501, detail="Not implemented")
+async def get_source(
+    hash: str, backend: Annotated[MatchboxDBAdapter, Depends(get_backend)]
+) -> dict[str, SourceItem] | str:
+    datasets = backend.datasets.list()
+    for dataset in datasets:
+        resolution = hexlify(dataset.resolution).decode("ascii")
```
lmazz1-dbt, line 118: can we use base64?

src/matchbox/server/api.py
```diff
@@ -76,18 +92,49 @@ async def clear_backend():
...
+        if resolution == hash:
+            result_obj = SourceItem(
+                table=dataset.table,
+                id=dataset.id,
+                schema=dataset.schema,
+                resolution=resolution,
+            )
+            return {"source": result_obj}
+    return "Source not found"
```
lmazz1-dbt, line 127: if the source is not found we need to return a 404. And it is found, let's just return the object, no need to wrap it in a dict.

src/matchbox/server/api.py
```diff
@@ -76,18 +92,49 @@ async def clear_backend():
...
+
+
+@app.post("/sources/uploadFile")
```
lmazz1-dbt, line 130: camelCase? Are we sure?

src/matchbox/server/api.py
```diff
@@ -76,18 +92,49 @@ async def clear_backend():
...
+async def add_source_to_s3(
+    file: UploadFile, bucket_name: str = Form(...), object_name: str = Form(...)
```
lmazz1-dbt, line 132: we shouldn't allow specifying a bucket name

src/matchbox/server/api.py
```diff
@@ -76,18 +92,49 @@ async def clear_backend():
...
+):
+    is_file_uploaded = upload_to_s3(file.file, bucket_name, object_name)
+    if is_file_uploaded:
+        return "File was successfully uplaoded"
+    return "File could not be uplaoded"
```
lmazz1-dbt, line 137: HTTP status codes please, and Pydantic schemas for the response

test/server/test_api.py
```diff
@@ -51,13 +52,40 @@ def test_count_backend_item(self, get_backend):
     #     response = client.post("/testing/clear")
     #     assert response.status_code == 200
 
-    # def test_list_sources():
-    #     response = client.get("/sources")
-    #     assert response.status_code == 200
+    @patch("matchbox.server.base.BackendManager.get_backend")
```
lmazz1-dbt, line 55: The tests will need to be updated to reflect requested changes in the API

lmazz1-dbt (changes requested):
As per my comments

PR #25, opened 2024-12-11
Sped up hierarchical clustering algorithm by 60%

ghost:
# Context

When creating dummy data I needed `_cluster_results_to_hierarchical` to run an order of magnitude faster. Adding this logic to the core repo feels like a free win. [Ticket](https://uktrade.atlassian.net/jira/software/projects/MTCHBX/boards/1123?selectedIssue=MTCHBX-31).

* Tracks ultimate parents in a way that means expensive graph traversal is no longer required

## Guidance to review

* Read the algo, confirm unit tests pass

## Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,6 @@ scratch/
 .ruff_cache
 notebooks/
 *.parquet
-.DS_Store
 
 # DuckDB
 *.duckdb


.gitignore
```diff
@@ -6,6 +6,9 @@ datasets.toml
 scratch/
 *.ipynb_checkpoints
 .ruff_cache
+notebooks/
+*.parquet
+.DS_Store
```
lmazz1-dbt, line 11: This is now duplicated from a previous PR

src/matchbox/server/postgresql/utils/insert.py
```diff
@@ -207,67 +207,59 @@ def _cluster_results_to_hierarchical(
...
+            current_ultimate_parents: set[bytes] = set()
+            for child in children:
+                current_ultimate_parents.update(ultimate_parents[child])
+
+            for up in current_ultimate_parents:
                 hierarchy.append((parent, up, threshold_float))
-                adj_dict[up].add((parent, threshold_float))
 
+            for child in children:
+                ultimate_parents[child] = {parent}
```
lmazz1-dbt, line 260: Clever - so unlike in my original suggestion, you're allowing multiple ultimate parents for each node: they are added at the point of adding new 2-node clusters because it's the only way this situation can arise. However, once we merge everything under a single cluster we're again in the world were there must be only one ultimate parent.

I still think there is no reason, [in this example](https://github.com/uktrade/matchbox/pull/20#discussion_r1871161117), to create an edge between `cvtu` and `zv`. But this will have a negligible impact on time now. It has more of an impact on space. I again think there is no good reason to create that edge on the basis that the hierarchical representation needs to be able to reconstruct:
- the probabilities, which it always can because we're storing 2-node clusters at the beginning of the process
- the clusters at any probability threshold, which it can do, because the only requirement is that there exists a path from a point in the hierarchy to all the leaves, which is the case even without the redundancy

lmazz1-dbt (changes requested):
As in my comments

.gitignore
lmazz1-dbt: But there is... at the bottom of gitignore in main

PR #33, opened 2024-12-20
Refactor structure into client/server/common

ghost:
# Context

While this PR touches dozens of files, it's intended to enact some very quick, uncontroversial changes that take very little approval. Controversial decisions have deliberately been de-scoped.

We've been working in a client/server/common format for some time but `src/` hasn't reflected this. We also know that `Results` has been in the wrong place. This fixes all that.

Not done in this PR is renaming `matchbox.models` to something like `matchbox.methodology`. It felt like an obvious change but as soon as I started doing it, it felt very wrong, because:

* This particular concept of "models" touches client side, server side, _and_ how we talk about these entities, and it's sometimes appropriate
* You end up with imports like `from matchbox.methodology.model import Model` and the renaming of those final objects in the path didn't seem as natural -- better to be out of scope for now

* `matchbox.helpers` moved to `matchbox.client.helpers`
* `matchbox.models` moved to `matchbox.client.models`
* `matchbox.clean` moved to `matchbox.client.clean`
* `matchbox.admin` moved to `matchbox.client.admin`
* `matchbox.common.results` split
    * `*Results` classes moved to `matchbox.client.results`
    * Transformation functions moved to `matchbox.common.transform` as they're about to principally be part of E*T*L operations

## Guidance to review

* Confirm you're happy with my moves
* Confirm `transform.py` was the right choice -- I could have done `hierarchy.py` but felt this was too specific. This leaves us open to `matchbox.common.transform.*` if this part of Matchbox grows

## Checklist:

### DIFF
--- a/src/matchbox/client/models/models.py
+++ b/src/matchbox/client/models/models.py
@@ -5,15 +5,15 @@ from pandas import DataFrame
 
 from matchbox.client.models.dedupers.base import Deduper
 from matchbox.client.models.linkers.base import Linker
-from matchbox.common.exceptions import MatchboxResolutionError
-from matchbox.common.results import (
+from matchbox.client.results import (
     ClusterResults,
     ModelMetadata,
     ModelType,
     ProbabilityResults,
     Results,
-    to_clusters,
 )
+from matchbox.common.exceptions import MatchboxResolutionError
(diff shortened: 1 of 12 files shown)


lmazz1-dbt (approved):
LGTM

PR #34, opened 2024-12-20
Connect `to_hierarchical_clusters` into the process with new `Results` and `insert_results`

ghost:
# Context

It's time to hook up all the different parts. This PR brings all our new client-side changes together.

This isn't intended to be super efficient yet -- just to hook things up.

* `Results` replaces `ProbabilityResults` and `ClusterResults`
    * `to_clusters` now just runs client-side and uses `IntMap` to produce its pseudo-hashes
* `insert_results` refactored to deal with the new structure
* `Probabilities.probability` changes to `SMALLINT` to match `pa.uint8()` on the frontend
* Adds `matchbox.common.logging` for console logging with `rich`, but doesn't move logger instantiation there (yet)
* Replaces all `pa.large_binary()` with `pa.binary()`. [Arrow advises against using the large version for our use case](https://arrow.apache.org/docs/python/generated/pyarrow.large_binary.html)

## Guidance to review

* Review `Results`
* Review `insert_results`
* Review everything else

## Relevant links

* [What we're aiming for](https://uktrade.atlassian.net/wiki/spaces/CDL/whiteboard/4434100236?atl_f=PAGETREE)

## Checklist:

### DIFF
--- a/test/server/test_postgresql.py
+++ b/test/server/test_postgresql.py
@@ -1,10 +1,13 @@
 from typing import Iterable
 
+import pyarrow as pa
+import pytest
 from sqlalchemy import text
 
 from matchbox.server.postgresql.benchmark.generate_tables import generate_all_tables
 from matchbox.server.postgresql.benchmark.init_schema import create_tables, empty_schema
 from matchbox.server.postgresql.db import MBDB
+from matchbox.server.postgresql.utils.insert import HashIDMap
 
 
 def test_benchmark_init_schema():
@@ -56,3 +59,40 @@ def test_benchmark_generate_tables():
                 if df[c].dtype == "uint64":
                     df[c] = df[c].astype("int64")
             df.to_sql(name=table_name, con=con, schema=schema)
+
+
+def test_hash_id_map():
+    """Test HashIDMap core functionality including basic operations."""
+    # Initialize with some existing mappings
+    lookup = pa.Table.from_arrays(
+        [
+            pa.array([1, 2], type=pa.uint64()),
+            pa.array([b"hash1", b"hash2"], type=pa.binary()),
+        ],
+        names=["id", "hash"],
+    )
+    hash_map = HashIDMap(start=100, lookup=lookup)
+
+    # Test getting existing hashes
+    ids = pa.array([2, 1], type=pa.uint64())
+    hashes = hash_map.get_hashes(ids)
+    assert hashes.to_pylist() == [b"hash2", b"hash1"]
+
+    # Test getting mix of existing and new hashes
+    input_hashes = pa.array([b"hash1", b"new_hash", b"hash2"], type=pa.binary())
+    returned_ids = hash_map.get_ids(input_hashes)
+
+    # Verify results
+    id_list = returned_ids.to_pylist()
+    assert id_list[0] == 1  # Existing hash1
+    assert id_list[2] == 2  # Existing hash2
+    assert id_list[1] == 100  # New hash got next available ID
+
+    # Verify lookup table was updated correctly
+    assert hash_map.lookup.shape == (3, 3)
+    assert hash_map.next_int == 101
+
+    # Test error handling for missing IDs
+    with pytest.raises(ValueError) as exc_info:
+        hash_map.get_hashes(pa.array([999], type=pa.uint64()))
+    assert "not found in lookup table" in str(exc_info.value)


src/matchbox/common/hash.py
```diff
@@ -70,13 +70,13 @@ def hash_data(data: str) -> bytes:
     return HASH_FUNC(prep_for_hash(data)).digest()
 
 
-def list_to_value_ordered_hash(list_: list[T]) -> bytes:
-    """Returns a single hash of a list ordered by its values.
+def hash_values(*values: *tuple[T, ...]) -> bytes:
```
leo-mazzone, line 73: T should be Comparable:
https://stackoverflow.com/a/37669538

src/matchbox/common/logging.py
```diff
@@ -0,0 +1,30 @@
...
+)
+
+
+def get_console():
+    """Get the console instance."""
+    return Console()
+
+
+def build_progress_bar(console: Console | None = None) -> Progress:
+    """Create a progress bar with custom columns."""
```
leo-mazzone, line 18: custom? They're hard-coded below

src/matchbox/server/postgresql/utils/insert.py
```diff
@@ -222,77 +223,235 @@ def _create_closure_entries(parent_resolution: Resolutions) -> None:
...
+                * cluster_hash, pa.large_binary()
+                * dataset, pa.uint64() (all Null)
+                * source_pk, pa.list_(type=pa.string())) (all Null)
+            * The Contains table
+                * parent, pa.uint64()
+                * child, pa.uint64()
+            * The Probabilities table
+                * resolution, pa.uint64()
+                * cluster, pa.uint64()
+                * probabilities, pa.uint8()
```
leo-mazzone, line 340: not sure how I feel about specifying the column name and types here. It will be difficult to spot it as a place to change if the columns change

src/matchbox/server/postgresql/utils/insert.py
```diff
@@ -385,70 +508,45 @@ def insert_results(
...
-                    hierarchy_unique_parents["threshold"],
-                    strict=True,
-                )
+                f"[{resolution.name}] Inserting {clusters.shape[0]:,} results "
+                "objects"
             )
 
             batch_ingest(
-                records=cluster_records,
+                records=[tuple(c.values()) for c in clusters.to_pylist()],
```
leo-mazzone, line 517: what guarantees that the columns are in the right order?

src/matchbox/server/postgresql/utils/insert.py
```diff
@@ -222,77 +223,235 @@ def _create_closure_entries(parent_resolution: Resolutions) -> None:
...
+                    ResolutionFrom.child == resolution_id
+                )
             )
+        )
+        .where(ResolutionFrom.child != resolution_id)
+    )
+
+    resolution_set = union(
+        direct_resolution, parent_resolutions, sibling_resolutions
+    ).cte("resolution_set")
```
leo-mazzone, line 285: How does this work? You're unioning results from Resolutions and ResolutionFrom tables?

src/matchbox/server/postgresql/utils/insert.py
```diff
@@ -222,77 +223,235 @@ def _create_closure_entries(parent_resolution: Resolutions) -> None:
...
+    )
 
-        # Process clusters at this threshold
-        current_clusters = cluster_df[cluster_df["threshold"] == threshold_float]
+    # Subquery for source datasets
+    source_datasets = (
+        select(Resolutions.resolution_id)
+        .join(
+            resolution_set, resolution_set.c.resolution_id == Resolutions.resolution_id
+        )
```
leo-mazzone, line 300: Why do you need to re-select from Resolutions?

src/matchbox/common/hash.py
leo-mazzone: fiar play on Comparable, I think I thought it was in `typing` but I must have misread. On the tuple, I meant the star symbol, so what I meant was, should it instead be:
```
def foo (*args: tuple[T,...]): ...
``` 
when it currently is:
```
def foo (*args: *tuple[T,...]): ...
```

src/matchbox/common/transform.py
```diff
@@ -256,10 +256,14 @@ def component_to_hierarchy(
             if len(children) <= 2:
                 continue  # Skip pairs already handled by pairwise probabilities
 
-            if im.has_mapping(*children):
-                continue  # Skip unchanged components from previous thresholds
+            # Skip if we've seen this exact component before
+            frozen_children = frozenset(children)
+            if frozen_children in seen_components:
+                continue
```
leo-mazzone, line 262: should we then remove the `has_mapping` method from `IntMap`?

src/matchbox/common/transform.py
```diff
@@ -34,37 +21,43 @@
 load_dotenv(dotenv_path)
 
 
-def to_clusters(results: ProbabilityResults) -> ClusterResults:
+def to_clusters(
```
leo-mazzone, line 24: No arg description in the docstring, which I think is important now that results is a generinc table rather than a Pydantic model

src/matchbox/common/transform.py
```diff
@@ -34,37 +21,43 @@
...
-            {"left_id": "large_binary[pyarrow]", "right_id": "large_binary[pyarrow]"}
-        )
-    )
+    # Sort probabilities descending and select relevant columns
+    results = results.sort_by([("probability", "descending")])
+    results = results.select(["left_id", "right_id", "probability"])
 
     # Get unique probability thresholds, sorted
-    thresholds = sorted(edges_df["probability"].unique())
+    thresholds = pc.unique(results.column("probability")).sort(order="descending")
```
leo-mazzone, line 44: you've already sorted above?

src/matchbox/common/transform.py
```diff
@@ -34,37 +21,43 @@
 load_dotenv(dotenv_path)
 
 
-def to_clusters(results: ProbabilityResults) -> ClusterResults:
+def to_clusters(
+    results: pa.Table,
+    dtype: pa.DataType = pa.large_binary,
```
leo-mazzone, line 26: - why is the default of the dtype large_binary, are we not working with ints?
- would it be clearer if called `id_dtype`?
- why are we making this settable anyway?

src/matchbox/common/transform.py
```diff
@@ -212,7 +210,9 @@ def get_components(self) -> list[set[T]]:
 
 
 def component_to_hierarchy(
-    table: pa.Table, salt: int, dtype: pa.DataType = pa.uint64
+    table: pa.Table,
+    dtype: pa.DataType = pa.large_binary,
+    hash_func: Callable[[*tuple[T, ...]], T] = hash_values,
```
leo-mazzone, line 215: The docstring needs to be updated

src/matchbox/common/transform.py
```diff
@@ -281,7 +285,8 @@ def component_to_hierarchy(
 def to_hierarchical_clusters(
     probabilities: pa.Table,
     proc_func: Callable[[pa.Table, pa.DataType], pa.Table] = component_to_hierarchy,
-    dtype: pa.DataType = pa.int64,
+    hash_func: Callable[[*tuple[T, ...]], T] = hash_values,
```
leo-mazzone, line 288: docstring

test/server/test_postgresql.py
```diff
@@ -1,263 +0,0 @@
-from typing import Iterable
```
leo-mazzone, line 1: Re-add

src/matchbox/client/results.py
```diff
@@ -43,263 +45,204 @@ class ModelMetadata(BaseModel):
...
-    @abstractmethod
-    def inspect_with_source(self) -> DataFrame:
-        """Enriches the results with the source data."""
-        return
+    @wraps(func)
+    def wrapper(self: "Results", *args: P.args, **kwargs: P.kwargs) -> R:
+        if not self.clusters:
+            im = IntMap(salt=42)
+            self.clusters = to_clusters(
+                results=self.probabilities, dtype=pa.int64, hash_func=im.index
```
leo-mazzone, line 56: Can you explain in a comment why you're using `IntMap` here instead of a proper hash function? I imagine the point is that this is meant to be used client-side but it's not clear

src/matchbox/client/results.py
```diff
@@ -43,263 +45,204 @@ class ModelMetadata(BaseModel):
...
+
+        if pa.types.is_floating(value["probability"].type):
+            if _check_range(value["probability"], 0, 1):
+                mult = 100
+            elif _check_range(value["probability"], 0, 100):
+                mult = 0
+            elif len(value["probability"]) == 0:
+                mult = 0
+                # Empty array, no need to check range
+                pass
```
leo-mazzone, line 118: pass?

src/matchbox/client/results.py
```diff
@@ -43,263 +45,204 @@ class ModelMetadata(BaseModel):
...
+            else:
+                p_max = pc.max(value["probability"])
+                p_min = pc.min(value["probability"])
+                raise ValueError(f"Probability range misconfigured: [{p_min}, {p_max}]")
+
+            probability_uint8 = pc.cast(
+                pc.multiply(value["probability"], mult),
```
leo-mazzone, line 125: you're multiplying probabilities between 0 and 1... by 0?
Even if that weren't an issue, I don't feel too good semantically about allowing floating points implicitly. Someone's probabilities might be between 0 and 1 on a 0-100 scale because the model is misconfigured. I think, if you're passing floats and your values are larger than 1, you get a validation error

src/matchbox/server/postgresql/utils/insert.py
leo-mazzone: I'd like to push back on your pushback 😄

src/matchbox/server/postgresql/utils/results.py
```diff
@@ -84,11 +77,9 @@ def _get_source_info(engine: Engine, resolution_id: int) -> SourceInfo:
     )
 
 
-def get_model_probabilities(
-    engine: Engine, resolution: Resolutions
-) -> ProbabilityResults:
+def get_model_results(engine: Engine, resolution: Resolutions) -> Results:
     """
-    Recover the model's ProbabilityResults.
+    Recover the model's Results, principally its pairwise probabilties.
```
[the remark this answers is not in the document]
leo-mazzone: Yeah, I think that the bit after "principally" is more confusing than helpful in this docstring, that's just my opinion though

leo-mazzone (approved):

PR #67, opened 2025-02-10
Removes the ModelAdapter ABC and adds a model_factory function

wpfl-dbt:
# Context

In order to implement API endpoints for models and results, a few things needed to happen:

* A factory for dummy models and results
* Some disentangling of what belongs to client, server and common for models and results
* Simplifying the two adapter ABCs into one

This PR aims to lay the groundwork for the API endpoint work to come. It also should help unblock #64, which is suffering for the client/server ambiguity of results

* `MatchboxModelAdapter` (and its PostgreSQL implementation) merged to `MatchboxDBAdapter`
* `Results` moved to client-side, `ModelMetadata` and `ModelType` to common DTOs
    * This also provoked a repo-wide standardisation of `left`/`right` to `left_id`/`right_id`, which were always the standard before my spike on hierarchies, where my draft versions of the names somehow ended up in the implemented code
    * Also `left/right_source` are now `left/right_resolution`
* `matchbox.common.factories.results` changed to `..models`, and simple `model_factory` built and tested
    * `generate_dummy_probabilities` gains a seed so it becomes reproducible
    * I'm anticipating `ModelMetrics` to need more attributes, it's just not integrated with the tests yet
* Adds PyArrow stubs as optional dependencies so `pa.Tables` are marginally easier to work with

## Guidance to review

I appreciate this could be split into two PRs, the factory and the adapter merge. Happy to do this if necessary, but I think the factory is simple and discrete enough not to add much complexity.

### Adapter change

* Review the new adapter ABC
* Review the new PostgreSQL adapter logic
* Review the changes to `Model` and `Results`
* Review the adapter unit tests

### Factory

* Review the factories
* Review the unit tests

## Checklist:


src/matchbox/server/postgresql/utils/insert.py
```diff
@@ -479,7 +478,7 @@ def insert_results(
     Args:
         resolution: Resolution of type model to associate results with
         engine: SQLAlchemy engine instance
-        results: A results object
+        results: A PyArrow results table left_id, right_id, probability
```
leo-mazzone, line 481: table _with_?

src/matchbox/common/dtos.py
```diff
@@ -38,6 +38,23 @@ def schema(self):
...
+
+
+class ModelMetadata(BaseModel):
+    """Metadata for a model."""
+
+    name: str
+    description: str
+    type: ModelType
+    left_source: str
+    right_source: str | None = None  # Only used for linker models
```
leo-mazzone, line 55: while you're refactoring... are these perhaps not quite the right names? "Source" is now strongly connotated in the codebase, and reusing it for something that could be another model feels like semantic overloading

test/common/test_factories.py
```diff
@@ -454,3 +458,81 @@ def test_source_factory_mock_properties():
...
+    model = model_factory(type=model_type)
+
+    assert model.model.type == model_type
+    assert (model.model.right_source is not None) == should_have_right_source
+
+    if model_type == ModelType.LINKER:
+        # Check that left and right values are in different ranges
+        left_vals = model.data.column("left_id").to_pylist()
+        right_vals = model.data.column("right_id").to_pylist()
+        assert all(lv < rv for lv, rv in zip(left_vals, right_vals, strict=False))
```
leo-mazzone, line 517: Nitpicking, but I always prefer to use strict=True once it's added by ruff

leo-mazzone: Also, what is that inequality actually testing?

test/common/test_factories.py
```diff
@@ -454,3 +458,81 @@ def test_source_factory_mock_properties():
...
+        pytest.param("deduper", False, id="deduper"),
+        pytest.param("linker", True, id="linker"),
+    ],
+)
+def test_model_factory_different_types(model_type: str, should_have_right_source: bool):
+    """Test model_factory handles different model types correctly."""
+    model = model_factory(type=model_type)
+
+    assert model.model.type == model_type
+    assert (model.model.right_source is not None) == should_have_right_source
```
leo-mazzone, line 511: or you could get rid of the variable and do
```python
if model_type == "linker`:
    assert model.model.right_source is not None
```

test/common/test_factories.py
```diff
@@ -454,3 +458,81 @@ def test_source_factory_mock_properties():
...
+    model1 = model_factory(seed=seed1)
+    model2 = model_factory(seed=seed2)
+
+    if should_be_equal:
+        assert model1.model.name == model2.model.name
+        assert model1.model.description == model2.model.description
+        assert model1.data.equals(model2.data)
+    else:
+        assert model1.model.name != model2.model.name
+        assert not model1.data.equals(model2.data)
```
leo-mazzone, line 538: what about the description if the seeds are different?

src/matchbox/server/postgresql/utils/results.py
```diff
@@ -104,86 +92,105 @@ def get_model_results(engine: Engine, resolution: Resolutions) -> Results:
...
+    # Final select to properly split out left and right
+    final_select = select(
+        pairs.c.id,
+        pairs.c.children[1].label("left_id"),
+        pairs.c.children[2].label("right_id"),
+        pairs.c.probability,
+    )
 
-        return Results(probabilities=df, metadata=metadata)
+    return sql_to_df(stmt=final_select, engine=engine, return_type="arrow")
```
leo-mazzone, line 196: This diff is tricky to follow, but I'm trusting that nothing has changed in terms of the logic?

src/matchbox/common/factories/models.py
```diff
@@ -287,5 +294,77 @@ def generate_dummy_probabilities(
...
+        description=description or generator.sentence(),
+        type=type,
+        left_source=generator.word(),
+        right_source=generator.word() if type == ModelType.LINKER else None,
+    )
+
+    left_values = range(n_true_entities)
+    right_values = None
+
+    if not type or ModelType.LINKER:
```
leo-mazzone, line 355: you could initialise type above, so it would also be part of the model metadata? Also, are you overriding a python keyword?

pyproject.toml
```diff
@@ -57,6 +57,7 @@ dev = [
 typing = [
     "boto3-stubs[s3]>=1.35.99",
     "polars>=1.11.0",
+    "pyarrow-stubs>=17.16",
```
leo-mazzone, line 60: Can you link an example of where this is used?
It seems fairly... safe based on the [Snyk report](https://snyk.io/advisor/python/pyarrow-stubs), but it doesn't even have a README?

test/common/test_factories.py
```diff
@@ -454,3 +458,87 @@ def test_source_factory_mock_properties():
...
+
+    if model_type == ModelType.LINKER:
+        assert model.model.right_resolution is not None
+
+        # Check that left and right values are in different ranges
+        left_vals = model.data.column("left_id").to_pylist()
+        right_vals = model.data.column("right_id").to_pylist()
+        left_min, left_max = min(left_vals), max(left_vals)
+        right_min, right_max = min(right_vals), max(right_vals)
+        assert (left_min < left_max <= right_min < right_max) or (
```
leo-mazzone, line 520: I don't think <= is right, cause that would mean you have the same value left and right

leo-mazzone (approved):

PR #71, opened 2025-02-18
Add factory for systems of linked sources

ghost:
# Context

Our factories need to cover the following needs:

* ✅ Sources with predictable outcomes to 
    * Test dedupers
    * Test helpers
    * Create controlled test scenarios for backend adapters to insert and test
    * Mock the API for the handler
* ✅ Models with predictable outcomes to
    * Create controlled test scenarios for backend adapters to insert and test
    * Mock the API for the handler
* ❌ Produce combinations of sources with predictable outcomes to 
    * Test linkers
    * Create controlled test scenarios for backend adapters to insert and test

What particularly highlighted the final set of needs was trying to set up tests for the `match()` function. The only way to hack it together with the existing factories, which don't track entities across sources, is by creating some very fragile fixtures. Tracking entities across sources also unlocks a much more robust way to check dedupers and linkers based on set membership, rather than probability counts -- way better.

This PR is to add a factory function to produce data for the third set of uses.

> [!WARNING]
> This is a big, complex PR, BUT the core of it is standalone scripts that are completely new, as are their unit tests. This should make dealing with it in one chunk less daunting than 4k lines of code would normally be.
> 
> I'm happy to break it into separate PRs if desired.

## Changes proposed in this pull request

* Removes `SourceMetrics` from the codebase
* Creates a standalone system of `SourceEntity` objects that are the "true" values across all sources
    * Adds an entity generator
    * Implements methods to search and reference entities across sources, and enrich with their raw data
* Adds `ResultsEntity` objects that are the "merging" values as they move through the system
* Refactors source factory and generator so `SourceEntity`s can be injected
* Adds `LinkedSourcesDummy` object that collects a system of sources and the entities they describe
* Refactors `ModelDummy`
    * Can now create predictably correct/incorrect sources
    * Can take `SourceDummy` or `ModelDummy`s as arguments
* Extensively unit tests the factory system
* Adds documentation to introduce the factory system 

## Guidance to review

* Read the documentation -- this is the simplest, best high level overview of this family of changes
* Read the new factories
* Read the new factory tests

## Relevant links

None.

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] New and existing unit tests pass locally with my changes
- [x] I've changed or updated relevant documentation

PR #79, opened 2025-02-28
Add an end to end test to show the entire system working

ghost:
# Context

#71 added a system of factories to help us reflect the client/server split in our testing. This is the first PR to use that system, and it uses it to implement our first proper end to end test.

* Adds an end to end test
    * Uses a new `e2e` schema for now to avoid conflict with legacy tests
* Adds a `docker` pytest mark to indicate tests that require resources in Docker
    * Updates the just command to allow `just test local`, `just test docker` and `just test`
* Adds a `query_to_cluster_entities()` function to create `ClusterEntities` for `query()`s from the real Matchbox to diff nicely with the factory system
* Fixed minor bug in the diff so that successful matches contribute to the overall mean match ratio
* Minor changes to client functions:
    * `select()` can suppress non-index warning
    * `trim()` added to cleaning steps

## Guidance to review

Confirm you're happy with the interim `e2e` schema solution while other tests migrate.

Other that reading and confirming the e2e test passes, consider bringing down all the Docker resources and running `just test local`.

## Checklist:

### DIFF
--- a/src/matchbox/client/helpers/selector.py
+++ b/src/matchbox/client/helpers/selector.py
@@ -34,7 +34,7 @@ def select(
             If not provided, will use a connection string from the
             `MB__CLIENT__DEFAULT_WAREHOUSE` environment variable.
         only_indexed: Whether you intend to select indexed columns only. Will raise a
-            warning if False and non-indexed columns are selected. Defaults to True.
+            warning if True and non-indexed columns are selected. Defaults to True.
             Non-indexed columns should only be selected if you're querying data for
             a purpose other than matching
 


src/matchbox/client/helpers/selector.py
```diff
@@ -24,14 +24,20 @@ class Selector(BaseModel):
...
     """From one engine, builds and verifies a list of selectors.
 
     Args:
-        selection: full source names and optionally a subset of columns to select
-        engine: the engine to connect to the data warehouse hosting the source.
+        selection: Full source names and optionally a subset of columns to select
+        engine: The engine to connect to the data warehouse hosting the source.
             If not provided, will use a connection string from the
             `MB__CLIENT__DEFAULT_WAREHOUSE` environment variable.
+        non_indexed: Whether you intend to select non-indexed columns. Will raise a
```
leo-mazzone, line 36: should it not... raise a warning if False? I think it would be more intuitive if this were a `only_indexed=True` argument

test/fixtures/db.py
```diff
@@ -386,12 +388,29 @@ def s3(aws_credentials: None) -> Generator[S3Client, None, None]:
...
-# Mock API
+# API, mocked and Docker
 
 
 @pytest.fixture(scope="function")
 def matchbox_api() -> Generator[MockRouter, None, None]:
+    """Client for the mocked Matchbox API."""
     with respx.mock(
-        base_url=client_settings.api_root, assert_all_called=True
+        base_url="http://localhost:8000", assert_all_called=True
```
leo-mazzone, line 398: why hardcode this? Doesn't seem like a good idea. It also makes it more difficult to redirect this to port 8080 when one wants to debug the API

test/common/factories/test_entity_factory.py
```diff
@@ -436,8 +470,7 @@ def test_source_to_results_conversion():
     assert results1.similarity_ratio(results3) == 0.0
 
     # Test error case for missing dataset
```
leo-mazzone, line 472: not a test error case

test/common/factories/test_entity_factory.py
```diff
@@ -346,6 +346,40 @@ def test_probabilities_to_results_entities(
...
+                make_cluster_entity(2, "d1", ["3", "4"]),  # Will be partially matched
+            ],
+            [
+                make_cluster_entity(1, "d1", ["1", "2"]),  # Same as in expected
+                make_cluster_entity(3, "d1", ["3", "5"]),  # Partial match with entity 2
+            ],
+            True,
+            False,
+            {
+                "mean_similarity": pytest.approx(5 / 9, rel=1e-2),
```
leo-mazzone, line 364: why 5 / 9?

src/matchbox/common/factories/entities.py
```diff
@@ -449,19 +450,62 @@ def to_cluster_entity(self, *names: str) -> ClusterEntity:
...
+        query: A PyArrow table or DataFrame representing a query result
+        source_pks: Mapping of source names to primary key column names
+
+    Returns:
+        A set of ClusterEntity objects
+    """
+    if isinstance(query, pa.Table):
+        query = query.to_pandas()
+
+    must_have_columns = set(["id"] + list(source_pks.values()))
```
leo-mazzone, line 486: multiple pks?

leo-mazzone: ah, sorry, one per source right?

test/e2e/test_e2e.py
```diff
@@ -0,0 +1,533 @@
...
+                full_name="e2e.cdms",
+                engine=warehouse_engine,
+                features=(
+                    features["crn"],
+                    features["cdms"],
+                ),
+                n_true_entities=n_true_entities,
+                repetition=1,  # Duplicate all rows
+            ),
+        )
```
leo-mazzone, line 104: I wonder whether this is a good time to start divorcing this from cdms and duns names? The former is an obscure legacy system no one remembers anymore, and the latter is a commercial product

test/e2e/test_e2e.py
```diff
@@ -0,0 +1,533 @@
...
+
+        # Use a separate schema to avoid conflict with legacy test data
+        # TODO: Remove once legacy tests are refactored
+        with warehouse_engine.connect() as conn:
+            conn.execute(text("create schema e2e;"))
+            conn.commit()
+
+        # Setup code - Create tables in warehouse
+        for source_name, source_testkit in self.linked_testkit.sources.items():
```
leo-mazzone, line 119: Will this be part of the later refactor? This is very much postgres-specific

test/e2e/test_e2e.py
```diff
@@ -0,0 +1,533 @@
...
+
+            return df
+
+        # === DEDUPLICATION PHASE ===
+        deduper_names = {}
+
+        for source_name, source_testkit in self.linked_testkit.sources.items():
+            # Get prefix for column names based on source
+            schema, table = source_name.replace('"', "").split(".")
+            prefix = f"{schema}_{table}_"
```
leo-mazzone, line 206: could you instead replace dots with underscores and add an underscore at the end? Not sure I've thought of all the edge cases, but would make this more general than a Postgres warehouse

test/e2e/test_e2e.py
```diff
@@ -0,0 +1,533 @@
...
+
+            # Query data from the source
+            # PK included then dropped to create ClusterEntity objects for later diff
+            source_select = select(
+                {
+                    source_name: ["pk"]
+                    + [col.alias for col in source_testkit.source.columns]
```
leo-mazzone, line 213: that's now what the alias is for I believe, you should use the name instead

test/e2e/test_e2e.py
```diff
@@ -0,0 +1,533 @@
...
+            select(
+                {duns_source: ["pk", "company_name", "duns"]},
+                engine=self.warehouse_engine,
+                non_indexed=True,
+            ),
+            select(
+                {cdms_source: ["pk", "crn", "cdms"]},
+                engine=self.warehouse_engine,
+                non_indexed=True,
+            ),
```
leo-mazzone, line 498: these are using the same engine, can all go in the same select statement

leo-mazzone (approved):
It's good to reap the benefits of your foundational work on testkits. This reads great, well  done. Incidentally, I could run all local tests (245) in 11 seconds.

PR #81, opened 2025-02-28
Remove docker marks from unit tests

ghost:
# Context

Some tests were using docker unnecessarily. This uses the factory system to remove this dependency unless it's actually needed.

`@pytest.mark.docker` now only appears in five places:

* Deduper tests (to be removed in separate PR)
* Linker tests (to be removed in separate PR)
* End to end tests
* Backend adapter tests
* Postgres benchmark helper tests

* `test_helpers` and `test_sources` updated
* Made it easier to write factory data to the warehouse
* Re-added `get_schema_table_names` to the codebase as it's useful

## Guidance to review

If you have a better way to deal with SQLite databases that need to be accessed via URIs in multiple places, I'd like to hear it. A named temp file seemed fine. The only other method I found required parameterising engine creation and I'd have had to do it within the codebase, not just the test.

## Relevant links

* [SQLAlchemy SQLite documentation on shared in-memory](https://docs.sqlalchemy.org/en/20/dialects/sqlite.html#using-a-memory-database-in-multiple-threads)

## Checklist:

### DIFF
--- a/test/client/test_helpers.py
+++ b/test/client/test_helpers.py
@@ -1,4 +1,3 @@
-import tempfile
 from typing import Callable
 from unittest.mock import Mock, patch
 
@@ -7,7 +6,7 @@ import pytest
 from httpx import Response
 from pandas import DataFrame
 from respx import MockRouter
-from sqlalchemy import create_engine
+from sqlalchemy import Engine
 
 from matchbox import index, match, process, query
 from matchbox.client.clean import company_name, company_number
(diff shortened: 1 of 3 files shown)


src/matchbox/common/factories/sources.py
```diff
@@ -103,9 +104,30 @@ def mock(self) -> Mock:
...
+        As the Source won't have an engine set by default, can be supplied.
+        """
+        schema, table = get_schema_table_names(self.name)
+        self.data.to_pandas().drop("id", axis=1).to_sql(
+            name=table,
+            schema=schema,
+            con=engine or self.source.engine,
+            index=False,
+            if_exists="replace",
+        )
```
leo-mazzone, line 129: what's this about? Schema+tablename won't work with sqlite

src/matchbox/common/factories/sources.py
```diff
@@ -535,7 +562,7 @@ def linked_sources_factory(
             source_configs = tuple(
                 SourceConfig(
                     full_name=config.full_name,
-                    engine=config.engine,
+                    engine=shared_engine if engine is not None else config.engine,
```
leo-mazzone, line 565:
```
engine if engine is not None else config.engine
```

Why inconveniencing `shared_engine`?

test/client/test_helpers.py
```diff
@@ -116,106 +113,96 @@ def test_select_missing_engine():
...
             con=conn,
-            schema="test",
             if_exists="replace",
             index=False,
         )
 
-        df.to_sql(
-            name="bar",
+        df_bar.to_sql(
+            name="cdms",
```
leo-mazzone, line 143: should your new `to_warehouse` method be used here?

test/client/test_helpers.py
```diff
@@ -116,106 +113,96 @@ def test_select_missing_engine():
...
-        db_pk="pk",
-        columns=[SourceColumn(name="a", type="BIGINT")],
-    )
-    source2 = Source(
-        address=SourceAddress.compose(engine=warehouse_engine, full_name="test.bar"),
-        db_pk="pk",
-    )
+def test_select_mixed_style(matchbox_api: MockRouter):
+    """We can select specific columns from some of the sources"""
+    engine = create_engine("sqlite:///:memory:")
```
leo-mazzone, line 118: I think we should be consistent, why are we using a tempfile above and here just memory?

test/client/test_helpers.py
leo-mazzone: same comment applies throughout

test/common/test_sources.py
```diff
@@ -67,48 +71,30 @@ def test_source_address_compose():
...
     # Error is raised with wrong type
     with pytest.raises(MatchboxSourceColumnError):
-        Source(
-            address=SourceAddress.compose(warehouse_engine, "test.foo"),
-            db_pk="pk",
-            columns=[
-                SourceColumn(name="b", type="BIGINT"),
-            ],
-        ).set_engine(warehouse_engine)
+        new_source = copy.copy(source_testkit.source)
```
leo-mazzone, line 95: would be better to distinguish among two errors, perhaps with some pytest.raises text matching?

test/client/test_helpers.py
leo-mazzone: also, could this be a fixture?

src/matchbox/common/factories/sources.py
leo-mazzone: Do we need the default engine at all? It's the same default from `SourceConfig`'s pydantic model

PR #82, opened 2025-03-03
Refactor methodology unit tests to use the factory system

ghost:
# Context

Our current tests for linkers and dedupers are:

* Too hard to understand, bringing in opaque fixtures from multiple places
* Too brittle, testing limited scenarios that aren't even stored with the tests
* Poorly scoped, testing extraneous features such as cleaning data and writing to the database

The factory system from #71 gives us the tools to fix this.

The current tests take three datasets, one with variations, one with missing data, and another with repetition (simple duplication), and run all methodologies over all of them after cleaning. Cleaning basically reduces the variations dataset to another repetition dataset. I therefore believe the relevant coverage from these tests ([dedupers](https://github.com/uktrade/matchbox/blob/main/test/client/test_dedupers.py) and [dedupe test params](https://github.com/uktrade/matchbox/blob/main/test/fixtures/models.py#L93), [linkers](https://github.com/uktrade/matchbox/blob/main/test/client/test_linkers.py) and [link test params](https://github.com/uktrade/matchbox/blob/main/test/fixtures/models.py#L141)) that needs replicating in this PR is:

For deduplication methodologies:

* ✅ Deterministic with no duplication
* ✅ Deterministic with simple duplication

For linking methodologies:

* ✅ Deterministic with no duplication
* ✅ Deterministic with simple duplication
* ✅ Deterministic with missing entities

All the above are accounted for. In addition to these I add:

* ✅ Deterministic deduplication with variations
* ✅ Deterministic linkage where nothing matches
* ✅ Probabilistic linkage with variations

Missing from our coverage, but that would require touching more than I think is appropriate for a refactoring PR, is:

* ❌ Probabilistic deduplication (because we don't yet implement it)
* ❌ Array-based logic
* ❌ Linkers and dedupers in different configurations in the matching DAG, enabled by `ModelTestkit`

This refactor shows where our coverage is missing, and demonstrates how our factory framework allows us to build towards covering it.

* Totally refactors `test_dedupers` and `test_linkers` into a series of smaller tests of concrete situations
* Implements the parameterisation of these tests in a way that can be read alongside the tests

There's also some minor changes that cropped up while building these tests:

* Some old typing patterns removed: `dict` for `typing.Dict`, `|` for `typing.Optional` etc (both fine since 3.10)
* Cleaning function docstrings got a tidy while I was in there
* Validation of probabilities in `Results` got a minor refactor to deal with empty results

## Guidance to review

It's important that you're happy that my test coverage is similar or higher than the tests they're replacing. I believe it is.

I think the high level structuring of tests is appropriate to contest here. I've done what I thought was right, and if you disagree, I welcome the challenge.

My recommended order of reading is:

* Deduper tests
* Linker tests
* All the minor changes

## Relevant links

* Original tests for [dedupers](https://github.com/uktrade/matchbox/blob/main/test/client/test_dedupers.py) and their [dedupe test params](https://github.com/uktrade/matchbox/blob/main/test/fixtures/models.py#L93)
* Original tests for [linkers](https://github.com/uktrade/matchbox/blob/main/test/client/test_linkers.py) and their [linker test params](https://github.com/uktrade/matchbox/blob/main/test/fixtures/models.py#L141)
* [PIP585](https://peps.python.org/pep-0585/) for type hinting generics
* [PEP604](https://peps.python.org/pep-0604/) for union typing

## Checklist:

### DIFF
--- a/test/client/methodologies/test_linkers_deterministic.py
+++ b/test/client/methodologies/test_linkers_deterministic.py
@@ -443,7 +443,4 @@ def test_no_matching_entities_linking(
     )
 
     assert not identical
-    # 10 perfect matches from the unlinked left_source, found in the linker
-    assert report["perfect"] == 10
-    # 10 invalid matches from the unlinked right_source, not in the linker
-    assert report["invalid"] == 10
+    assert results.probabilities.num_rows == 0


test/client/methodologies/test_dedupers_deterministic.py
```diff
@@ -0,0 +1,213 @@
+"""Test deterministic behavior of dedupers."""
```
leo-mazzone, line 1: Or rather, "test behaviour of deterministic dedupers"

test/client/methodologies/test_dedupers_deterministic.py
```diff
@@ -0,0 +1,213 @@
...
+    """Configure settings for NaiveDeduper.
+
+    Args:
+        source: Source object from linked_sources_factory
+
+    Returns:
+        A dictionary with validated settings for NaiveDeduper
+    """
+    # Extract column names excluding pk and id
+    fields = [c.name for c in source.source.columns if c.name not in ("pk", "id")]
```
leo-mazzone, line 34: What determines the hardcoding of pk and id?

test/client/methodologies/test_linkers_deterministic.py
```diff
@@ -0,0 +1,421 @@
...
+        model_settings=configure_linker(left_source, right_source),
+        left_data=left_source.query.to_pandas().drop("pk", axis=1),
+        left_resolution="source_left",
+        right_data=right_source.query.to_pandas().drop("pk", axis=1),
+        right_resolution="source_right",
+    )
+    results: Results = linker.run()
+
+    # Validate results against ground truth
+    identical, report = linked.diff_results(
```
leo-mazzone, line 250: Could you assert that the left and right are identical? To reassure me that you're testing the right thing, as it's obscured by the testkit setup

test/client/methodologies/test_linkers_deterministic.py
```diff
@@ -0,0 +1,421 @@
...
+            base_generator="email",
+        ),
+    )
+
+    configs = (
+        SourceConfig(
+            full_name="source_left",
+            features=features,
+            n_true_entities=10,
+            repetition=1,  # Each entity appears once
```
leo-mazzone, line 284: Either the comment or the param is wrong

test/client/methodologies/test_linkers_deterministic.py
```diff
@@ -0,0 +1,421 @@
...
+    """
+    # Create two completely separate sets of entities
+    features = (
+        FeatureConfig(name="company", base_generator="company"),
+        FeatureConfig(name="identifier", base_generator="uuid4"),
+    )
+
+    # Create the sources with disjoint entity sets
+    left_source = source_factory(features=features, n_true_entities=10, seed=314)
+    right_source = source_factory(features=features, n_true_entities=10, seed=159)
```
leo-mazzone, line 405: Having a different seed won't guarantee there are no overlaps... could you at least assert it? Or is there some other way to guarantee it?

test/client/methodologies/test_linkers_probabilistic.py
```diff
@@ -0,0 +1,242 @@
...
+        ):
+            blocking_rules.append(f"CAST(l.{field} AS INT) = CAST(r.{field} AS INT)")
+            comparisons.append(cl.ExactMatch(field))
+
+        else:
+            comparisons.append(cl.ExactMatch(field))
+
+    # Use simple modulo blocking if no blocking rules were created
+    if not blocking_rules:
+        blocking_rules = ["l.id % 10 = r.id % 10"]
```
leo-mazzone, line 123: why this instead of having no blocking rules? Presumably these are small examples, so this is not for efficiency, but is it for checking the behaviour of blocking rules?

test/client/methodologies/test_linkers_probabilistic.py
```diff
@@ -0,0 +1,242 @@
...
+        probabilities=results.probabilities,
+        left_clusters=left_source.entities,
+        right_clusters=right_source.entities,
+        sources=["source_left", "source_right"],
+        threshold=np.mean(results.probabilities["probability"]),
+        verbose=True,
+    )
+
+    assert not identical, f"Expected imperfect results but got: {report}"
+    assert len(report["missing"]) == 0
```
leo-mazzone, line 242: if there is nothing missing... does this mean that results include all records on the left and right, but not necessarily in the right clusters?

leo-mazzone: how can we be sure some of them won't be missing altogether, for example as a result of the mod 10 blocking rules?

test/client/methodologies/test_linkers_deterministic.py
```diff
@@ -0,0 +1,421 @@
...
+        model_settings=configure_linker(left_source, right_source),
+        left_data=left_source.query.to_pandas().drop("pk", axis=1),
+        left_resolution="source_left",
+        right_data=right_source.query.to_pandas().drop("pk", axis=1),
+        right_resolution="source_right",
+    )
+    results = linker.run()
+
+    # For this case, we expect no probabilities or empty results
+    assert results.probabilities.shape[0] == 0, "Expected no matching probabilities"
```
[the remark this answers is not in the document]
leo-mazzone: What do you think? If it's redundant, let's remove it

src/matchbox/common/factories/entities.py
```diff
@@ -582,99 +582,178 @@ def diff_results(
...
 
     Returns:
-        Tuple of (is_identical, diff_dict)
+        A tuple containing:
+        - Boolean: True if lists are identical, False otherwise
+        - Dictionary with comparison details:
+            - 'perfect_matches': Entities that match exactly in both lists
+            - 'fragmented_matches': Expected entities that are split into multiple
+                fragments in the actual results
+            - 'unexpected_matches': Actual entities that incorrectly merge multiple
```
leo-mazzone, line 591: multiple what?

src/matchbox/common/factories/entities.py
```diff
@@ -582,99 +582,178 @@ def diff_results(
         verbose: Whether to return detailed diff report
 
     Returns:
-        Tuple of (is_identical, diff_dict)
+        A tuple containing:
+        - Boolean: True if lists are identical, False otherwise
+        - Dictionary with comparison details:
+            - 'perfect_matches': Entities that match exactly in both lists
```
leo-mazzone, line 588: what's an entity? A cluster? A record?

src/matchbox/common/factories/sources.py
```diff
@@ -194,9 +194,24 @@ def diff_results(
...
 
         Returns:
-            A tuple of bool, dict:
-            * Whether the results match the true entities
-            * A dictionary report of differences
+            A tuple containing:
+            - Boolean: True if lists are identical, False otherwise
+            - Dictionary with comparison details:
+                - 'perfect_matches': Entities that match exactly in both lists
+                - 'fragmented_matches': Expected entities that are split into multiple
```
leo-mazzone, line 201: multiple...

src/matchbox/common/factories/sources.py
```diff
@@ -194,9 +194,24 @@ def diff_results(
             verbose: Whether to include verbose information in the report
 
         Returns:
-            A tuple of bool, dict:
-            * Whether the results match the true entities
-            * A dictionary report of differences
+            A tuple containing:
```
leo-mazzone, line 197: Is this identical to the docstring in the other file? Could you just point to that one instead so they don't get out of sync?

test/common/factories/test_model_factory.py
```diff
@@ -237,10 +237,11 @@ def test_model_pipeline_with_dummy_methodology(
 
     # Verify the imperfect methodology was detected
     assert not identical
-    assert report["mean_similarity"] < 1.0
     # Random process: can't guarantee particular problems, but can guarantee
     # that some will be present
-    assert any([report["partial"], report["missing"], report["extra"]])
+    assert any([report["fragmented_matches"], report["unexpected_matches"]])
```
leo-mazzone, line 242: if you are removing one entity intentionally, you should get fragmented, not unexpected, right?

src/matchbox/common/factories/entities.py
```diff
@@ -582,99 +582,178 @@ def diff_results(
...
+            for e in perfect_matches
+        ]
+        result["fragmented_matches"] = []
+        result["unexpected_matches"] = []
+        result["missing_matches"] = []
+        result["spurious_matches"] = []
+
+    # Process fragmented matches
+    for expected_entity in remaining_expected:
+        fragments = [
```
leo-mazzone, line 634: note to self: these are entities with partial matches

src/matchbox/common/factories/entities.py
```diff
@@ -582,99 +582,178 @@ def diff_results(
...
-
-    if not verbose:
-        return False, result
+            if any(overlap_pks.values()):
+                temp = ClusterEntity(
+                    source_pks=EntityReference(
+                        {k: v for k, v in overlap_pks.items() if v}
+                    )
+                )
+                combined = temp if combined is None else combined + temp
```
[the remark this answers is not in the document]
leo-mazzone: I posit: never

test/client/methodologies/test_linkers_probabilistic.py
leo-mazzone: We agreed to change the diff results to:
- Remove precision and recall
- Consider the following cluster-wise outcomes from comparing expected and actual:
image

- Making the diff logic much simpler

A few examples:
```
e: (1, 2) (3, 4), (5, 6)
a: (1, 2, 3, 4), (5, 6)
(1, 2, 3, 4) is a superset (of more than one expected cluster, but we don't care)
(5, 6 is perfect)


e: (1, 2) (3, 4), (5, 6)
a: (1, 6) (2, 3, 4, 5)
(1, 6) is "wrong"
(2, 3, 4, 5) is also "wrong"


e: (1, 2) (3, 4) (5, 6)
a: (1, 2) (3, 4) (5)
first two are perfect
the last one is a subset (even though 6 never appears in any actual cluster)

e: (1, 2)
a: (1, 2, 3)
(1, 2, 3) is invalid, because 3 is an unexpected PK
```

leo-mazzone (approved):
LGTM

PR #83, opened 2025-03-04
Refactor adapter tests

ghost:
# Context

The backend adapter tests currently:

* Use the API
* Take forever
* Rely on some very confusing fixtures
* Are rigid and brittle because they rely on a limited set of data in their tests

This PR is to bring the factory testkit system to backend adapter tests, making them faster and more useful.

* Adds a `TestkitDAG` to collect systems of testkits together into client-side representations of the database state
* Implements a snapshotting system to cache the database state and the testkit that goes with it
* Implements a new `setup_scenario` function to replace the existing setup system
* Adds `query_to_model_factory()` to form `ModelTestkit`s from queries to the real Matchbox
* Moves all tests over to the scenario and testkit system, including a fundamental rewrite of almost all the backend test functions
    * Uses SQLite warehouses instead of PostgreSQL in these tests
* Removed the cache from `generate_entity_probabilities()` as the way `ClusterEntity`s calculate equality (without the ID) meant that the cached values messed with the workings of `query_to_model_factory()`

## Guidance to review

Some files in this PR, like the backend adapter tests, have almost every line touched with a change. I'd read them as a new file rather than the diff.

I'd start with `query_to_model_factory()` and `TestkitDAG` to understand the building blocks of these changes.

## Relevant links

None yet.

## Checklist:

### DIFF
--- a/test/fixtures/db.py
+++ b/test/fixtures/db.py
@@ -71,6 +71,10 @@ def create_scenario_dag(
     This approach first creates source data, writes it to backend and warehouse,
     then builds models by querying the backend to ensure ID alignment.
     """
+    # Validate inputs
+    if scenario_type not in ["bare", "index", "dedupe", "link"]:
+        raise ValueError(f"Invalid scenario: {scenario_type}")
+
     dag = TestkitDAG()
 
     # 1. Create linked sources


pyproject.toml
```diff
@@ -103,6 +103,7 @@ line-ending = "auto"
 testpaths = ["test"]
 pythonpath = ["."]
 addopts = "-s -vv --cov=matchbox test/ --log-disable=pg_bulk_ingest"
+norecursedirs = "src"
```
leo-mazzone, line 106: what does this do, if the testpath is already outside of src?

src/matchbox/common/sources.py
```diff
@@ -109,12 +111,30 @@ class Source(BaseModel):
     db_pk: str
     columns: list[SourceColumn] = []
 
-    _engine: Engine | None = None
+    _engine: Engine | None = PrivateAttr(default=None)
```
leo-mazzone, line 114: Should this not happen automatically?
https://docs.pydantic.dev/latest/concepts/models/#private-model-attributes

leo-mazzone: yes I think so

test/server/test_postgresql.py
```diff
@@ -61,97 +62,98 @@ def test_hash_id_map():
...
-        source_pk=source_pks[0],
-        source_name=warehouse_data[1].address.full_name,  # DUNS
-        point_of_truth="deterministic_naive_test.crn_naive_test.duns",
-    )
+    with setup_scenario(matchbox_postgres, "link", warehouse=postgres_warehouse) as dag:
+        engine = MBDB.get_engine()
+
+        linker_name = "deterministic_naive_test.crn_naive_test.duns"
+        linked, _ = dag.get_sources_for_model(linker_name)
+        # crn_testkit = dag.sources.get("crn")
```
leo-mazzone, line 129: take out?

src/matchbox/common/factories/dags.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def _update_root_sources(self, model_name: str):
+        """Update root sources for a model."""
+        for dep_name in self.adjacency[model_name]:
+            if dep_name in self.sources:
+                # If dependency is a source, add it directly
+                self.root_sources[model_name].add(dep_name)
+            elif dep_name in self.models:
+                # If dependency is a model, add all its root sources
+                self.root_sources[model_name].update(self.root_sources[dep_name])
```
leo-mazzone, line 39: It's interesting how we really are doing similar things in my DAG and your test kit DAG. On my end, it will ensure when you set up a linking step and then you pass it a query based on a previous resolution, it will check whether the resolutions feeding into that step will have access to the sources you want

leo-mazzone: (feel free to resolve this comment)

src/matchbox/common/factories/dags.py
```diff
@@ -0,0 +1,110 @@
...
+        left_res = testkit.model.metadata.left_resolution
+        if left_res:
+            self.adjacency[model_name].add(left_res)
+
+        right_res = testkit.model.metadata.right_resolution
+        if right_res and right_res != left_res:
+            self.adjacency[model_name].add(right_res)
+
+        # Update root sources
+        self._update_root_sources(model_name)
```
leo-mazzone, line 71: what happens if you don't add the models in the right order?

src/matchbox/common/factories/dags.py
```diff
@@ -0,0 +1,110 @@
...
+        # Get the first linked source and its testkit
+        first_linked_source = next(iter(linked_sources))
+        linked_idx = self.source_to_linked[first_linked_source]
+        linked_testkit = self.linked[linked_idx]
+
+        # Get all sources from this linked testkit
+        linked_source_names = set(linked_testkit.sources.keys())
+
+        # Find which root sources belong to this linked testkit
+        model_sources = sorted(list(root_sources.intersection(linked_source_names)))
```
leo-mazzone, line 108: why the sorting?

src/matchbox/common/factories/dags.py
```diff
@@ -0,0 +1,110 @@
...
+        # Find root sources that have a linked testkit
+        linked_sources = {s for s in root_sources if s in self.source_to_linked}
+
+        if not linked_sources:
+            return None, []
+
+        # Get the first linked source and its testkit
+        first_linked_source = next(iter(linked_sources))
+        linked_idx = self.source_to_linked[first_linked_source]
+        linked_testkit = self.linked[linked_idx]
```
leo-mazzone, line 102: when would it be the case that you have some sources but no linked sources? Or more in general, a source without it being linked to anything?
Could there be more than one linked_sources and if so why are we just returning the first one?

src/matchbox/server/postgresql/utils/db.py
```diff
@@ -61,6 +70,123 @@ def get_resolution_graph(engine: Engine) -> ResolutionGraph:
...
+
+
+def restore(engine: Engine, snapshot: MatchboxSnapshot, batch_size: int) -> None:
+    """Restores the database from a snapshot.
+
+    Args:
+        snapshot: A MatchboxSnapshot object of type "postgres" with the
+            database's state
+        batch_size: The number of records to insert in each batch
+    """
```
leo-mazzone, line 131: are you testing restore and dump?

src/matchbox/server/postgresql/utils/db.py
```diff
@@ -61,6 +70,123 @@ def get_resolution_graph(engine: Engine) -> ResolutionGraph:
...
+                record_dict = {}
+                for column in inspect(model).columns:
+                    value = getattr(record, column.name)
+
+                    # Store bytes as base64
+                    if isinstance(value, bytes):
+                        value = base64.b64encode(value).decode("ascii")
+                        # Mark this as encoded so we can decode it on restoration
+                        record_dict[f"{column.name}_format"] = "base64"
+
```
leo-mazzone, line 114:
```
{foo: "bar", foo_type: "base64"}
```
is a bit weird. Perhaps the following instead?
```
{foo: {"base64": "bar"}}
```

test/common/factories/test_testkit_dag.py
```diff
@@ -0,0 +1,301 @@
...
+
+    This test verifies that:
+    1. We can create different types of models: single source dedupers, linkers, chains
+    2. The DAG correctly tracks model dependencies
+    3. We can find the original LinkedSourcesTestkit from any model in the chain
+    4. We can identify the correct sources for each model
+    5. Standalone sources are handled correctly
+    """
+    # Create linked sources
+    linked = linked_sources_factory(seed=hash(str(chain_config)) % 1000000)
```
leo-mazzone, line 137: what's that modulo doing?

test/common/factories/test_testkit_dag.py
```diff
@@ -0,0 +1,301 @@
...
+        # Set up left input (required for all models)
+        if previous_model_names:
+            left_testkit = models[previous_model_names[0]]
+        elif is_standalone:
+            left_testkit = standalone
+        elif source_names:
+            left_testkit = linked.sources[source_names[0]]
+
+        # Set up right input (optional, for linkers)
+        if previous_model_names and len(previous_model_names) >= 2:
```
leo-mazzone, line 176: when would it be greater than 2?

test/common/factories/test_testkit_dag.py
```diff
@@ -0,0 +1,301 @@
...
+            assert linked_testkit is None, (
+                f"Model {model_name} should not find linked sources testkit"
+            )
+        else:
+            assert linked_testkit is not None, (
+                f"Model {model_name} should find linked sources testkit"
+            )
+            assert linked_testkit is linked, (
+                f"Model {model_name} should find the original linked testkit"
+            )
```
leo-mazzone, line 216: these two the same?

test/common/factories/test_testkit_dag.py
```diff
@@ -0,0 +1,301 @@
...
+            left_testkit = standalone
+        elif source_names:
+            left_testkit = linked.sources[source_names[0]]
+
+        # Set up right input (optional, for linkers)
+        if previous_model_names and len(previous_model_names) >= 2:
+            right_testkit = models[previous_model_names[1]]
+        elif source_names and (
+            len(source_names) >= 2 or (previous_model_names and source_names)
+        ):
```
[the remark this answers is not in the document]
leo-mazzone: at the very least you don't need the nested `source_names is not None` condition

test/fixtures/db.py
```diff
@@ -1,320 +1,304 @@
...
-                _, df_l = request.getfixturevalue(fx_data.fixture_l)
-                _, df_r = request.getfixturevalue(fx_data.fixture_r)
+    # End here for index-only scenarios
+    if scenario_type == "index":
+        return dag
 
-                linker_name = f"{fx_linker.name}_{fx_data.source_l}_{fx_data.source_r}"
-                linker_settings = fx_linker.build_settings(fx_data)
+    # 4. Create and add deduplication models
+    if scenario_type in ["dedupe", "link"]:
```
leo-mazzone, line 101: it must be, or we would have returned

test/fixtures/db.py
```diff
@@ -1,320 +1,304 @@
...
+    backend.clear(certain=True)
+
+
+# Filesystem fixtures
 
 
 @pytest.fixture(scope="session")
-def warehouse_engine() -> Engine:
+def test_root_dir() -> Path:
+    return TEST_ROOT
```
leo-mazzone, line 284: does this belong in this file?

test/fixtures/db.py
```diff
@@ -1,320 +1,304 @@
...
-        AddIndexedDataCallable,
-        MatchboxPostgres,
-        list[Source],
-        list[DedupeTestParams],
-        list[ModelTestParams],
-        FixtureRequest,
-    ],
-    None,
-]
+def _reconcile_engine_and_testkitdag(engine: Engine, dag: TestkitDAG) -> None:
```
leo-mazzone, line 54: name's a bit confusing. Perhaps "_testkitdag_to_warehouse"

test/fixtures/db.py
leo-mazzone: Ah I see a problem with the above now: if you mispell dedupe, it will then assume you meant link

leo-mazzone (approved):
Really solid work

PR #91, opened 2025-03-17
Tidy up logging

ghost:
# Context

Builds on the insights generated in #85 to fix the noisy logger. Closes #85 

* Running `pytest` will collect test results by file and print no logs
* Running tests in VSCode will similarly collect tests results by file and print no longs
* Running debug in VSCode will run the tests one by one, suppress coverage, and print all logs
* Logs in GitHub Actions should collect by file and only print logs for errors

To turn logging on/up, `pyproject.toml` (or on your own CLI) is the only place to make a change.

* Changing to `log_cli = true`/ `log_cli_level = "DEBUG"` before you run tests will turn the live logs back on and show test results one by one
* Similarly, `pytest --log-cli-level="DEBUG"` will have the same effect

A big part of our problem was the `-s` flag printing logs whether we wanted them or not.

* `pyproject.toml` edited to control log levels
* All but essential options removed from `settings.json` and `launch.json`
* Cut some comments from centralised logging

## Guidance to review

Note that you can't turn off logs in debug mode -- to do that we'd need an application-level flag. This is because of the `--capture=no` flag VSCode's extension mandates.

* Run pytest
* Run test in VSCode
* Debug test in VSCode
* Change the options as I describe and see logs reappear

## Relevant links

* [pytest logging documentation](https://docs.pytest.org/en/stable/how-to/logging.html)

## Checklist:

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -11,14 +11,10 @@
             "justMyCode": false,
             "env": {
                 "PYTEST_ADDOPTS": "--no-cov",
-                "PYTHONPATH": "${workspaceFolder}"
+                "PYTHONPATH": "${workspaceFolder}",
             },
             "python": "${workspaceFolder}/.venv/bin/python",
             "cwd": "${workspaceFolder}",
-            "args": [
-                "-v",
-                "-s"
-            ]
         },
         {
             "name": "Matchbox: Profile",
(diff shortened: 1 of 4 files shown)


leo-mazzone (approved):
LGTM

PR #95, opened 2025-03-19
Implement batch reading from warehouses

ghost:
# Context

When we ran query and index on a smallish PostgreSQL warehouse with a bigish dataset, we brought it down. We need to be able to batch.

Unfortunately, [connectorx doesn't currently open up batch reading](https://github.com/pola-rs/polars/issues/21041) to its Python API. We considered `limit`/`offset` and decided this was too risky. While Polars' author has said they're [not an Arrow replacement](https://news.ycombinator.com/item?id=26454585), we've discussed at length that our codebase performs plenty of transformations better-implemented in Polars than PyArrow -- or certainly _our_ PyArrow. We're fine with making a switch, and make use of that decision here.

We therefore agreed to back `sql_to_df()` with Polars to open up batching.

One problem with this is that Polars' interpretation of Arrow's datatypes differs from ours. They default to `large_binary` over `binary`, `large_string` over `string`, and `large_list` over `list`. When it comes to strings, [they've thought very carefully about this](https://pola.rs/posts/polars-string-type/), and I trust them. Conversely, Arrow are very keen on `large_binary` [not being used](https://arrow.apache.org/docs/python/generated/pyarrow.large_binary.html), and I trust them too. This is made trickier in that converting from large to small isn't always easy/cheap.

One solution that won't work is [`pl.read_database(schema_overrides=)`](https://docs.pola.rs/api/python/stable/reference/api/polars.read_database.html#polars-read-database). It won't work because it needs Polars types, and Polars doesn't differentiate between [large and small binary](https://docs.pola.rs/api/python/stable/reference/api/polars.datatypes.Binary.html#polars.datatypes.Binary) (or large/small anything).

Right now, I've proposed going with Polars' types.

### Empirical testing of Polars typing on size

I generated two datasets with 100k true entities, looking like this:

| pk | id | company_name | crn |
|----|----|--------------|----|
| 790c76b2-3202-4169-ba4a-699839ba6629 | 9821148752815052 | "Daniel, Green and Diaz" | DDu-117-Prt-094 |
| 19ced616-0773-4e04-8e48-e9b15ae6c603 | 3555780734488527 | "Lopez, Jones and May" | Qvy-192-TMK-347 |
| f6b18c4d-c76f-4a92-8380-23d0948000be | 9533747396063057 | Hudson-Shaw | APZ-701-cbh-872 |
| 92ee3fb4-5279-44d2-b47e-b1c0e33051a4 | 9976824860132760 | Rodriguez-Howard | leZ-270-YmP-993 |
| 48efd766-a6f5-43b1-b08a-2a117cd976b9 | 8561239618809288 | Chan-Richardson | bAt-720-iUk-683 |

One had no ("low") repetition, the other had four primary keys for each ID ("high" repetition), which would lead to `source_pk` arrays.

I then generated the Arrow shape we send to the API for indexing such that we had the "large" and "small" size. Here's the schemas:

```python
SCHEMA_INDEX_LRG = pa.schema(
    [("hash", pa.large_binary()), ("source_pk", pa.large_list(pa.large_string()))]
)

SCHEMA_INDEX_SML = pa.schema(
    [("hash", pa.binary()), ("source_pk", pa.list_(pa.string()))]
)
```

I then measured their size in memory and on disk. I also unnested the arrays for the repeated datasets to see if the "large" types provided any savings.

Datatype size | Repetition | Nesting | Rows | Keys per row | Memory size | Disk size
-- | -- | -- | -- | -- | -- | --
Large | High | Nested | 100,000 | 4 | 22.80 MB | 8.02 MB
Small | High | Nested | 100,000 | 4 | 20.40 MB | 8.02 MB
Large | Low | Nested | 100,000 | 1 | 9.60 MB | 7.32 MB
Small | Low | Nested | 100,000 | 1 | 8.40 MB | 7.32 MB
Large | High | Unnested | 400,000 | - | 35.20 MB | 8.71 MB
Small | High | Unnested | 400,000 | - | 32.00 MB | 8.71 MB

I found:

* "Large" types are larger by about 1 MB (low repetition) - 2.5 MB (high repetition) / 100k in memory and nothing / 100k on disk
* Unnesting the strings reveals a similar difference, and unnested remains bigger
    * If there's saving on repetitions, it's in `large_string`, not `large_binary`
    * Bonus: when converting `large_binary` to base64 and encoding highly-repeated data as `large_string`, there's _still_ no saving, and I wonder if our hashes are too short to trigger the saving Polar talk about in the linked blog

> [!NOTE]
> The cost for using large datatypes is in memory alone, and that cost is pretty negligible. I believe it's worth paying so we can use the simpler, more flexible, higher level API of Polars throughout.

* Updates `sql_to_df()` to use Polars
* Updates `Source`'s `to_*()` functions to use them
* Updates Arrow transfer schemas to use Polars' chosen types

## Guidance to review

Polars and Arrow types: what do we do? Go with Polars, manually convert, or something else?

## Relevant links

* [Connectorx and batch reading](https://github.com/pola-rs/polars/issues/21041) on the Polars issue tracker
* [Polars and Arrow `large_string` type](https://pola.rs/posts/polars-string-type/)
* [Polars documentation on Arrow interoperability](https://docs.pola.rs/user-guide/misc/arrow/)
* [Polars `pl.Binary` type documentation](https://docs.pola.rs/api/python/stable/reference/api/polars.datatypes.Binary.html#polars.datatypes.Binary)
* [Arrow large binary documentation](https://arrow.apache.org/docs/python/generated/pyarrow.large_binary.html)

## Checklist:

### DIFF
--- a/src/matchbox/client/helpers/selector.py
+++ b/src/matchbox/client/helpers/selector.py
@@ -1,17 +1,20 @@
 """Functions to select and retrieve data from the Matchbox server."""
 
 import itertools
-from typing import Iterator, Literal
+from typing import Iterator, Literal, get_args
 from warnings import warn
 
+import polars as pl
 import pyarrow as pa
-from pandas import ArrowDtype, DataFrame
+from pandas import ArrowDtype
+from pyarrow import Table as ArrowTable
 from pyarrow import compute as pc
 from pydantic import BaseModel
 from sqlalchemy import Engine, create_engine
 
 from matchbox.client import _handler
 from matchbox.client._settings import settings
+from matchbox.common.db import QueryReturnType, ReturnTypeStr
 from matchbox.common.graph import DEFAULT_RESOLUTION
 from matchbox.common.logging import logger
 from matchbox.common.sources import Match, Source, SourceAddress
@@ -95,8 +98,8 @@ def select(
 
 
 def _process_query_result(
-    data: pa.Table, selector: Selector, mb_ids: pa.Table
-) -> pa.Table:
+    data: ArrowTable, selector: Selector, mb_ids: ArrowTable
+) -> ArrowTable:
     """Process query results by joining with matchbox IDs and filtering fields.
 
     Args:
@@ -129,9 +132,9 @@ def _query_batched(
     sub_limits: list[int | None],
     resolution_name: str | None,
     threshold: int | None,
-    return_type: Literal["pandas", "arrow"],
+    return_type: ReturnTypeStr,
     batch_size: int | None,
-) -> Iterator[DataFrame] | Iterator[pa.Table]:
+) -> Iterator[QueryReturnType]:
     """Helper function that implements batched query processing.
 
     Returns an iterator yielding batches of results.
@@ -176,27 +179,30 @@ def _query_batched(
 
     # Convert each batch to the requested return type
     for batch in batches_iter:
-        if return_type == "pandas":
-            yield batch.to_pandas(
-                use_threads=True,
-                split_blocks=True,
-                self_destruct=True,
-                types_mapper=ArrowDtype,
-            )
-        else:
-            yield batch
+        match return_type:
+            case "pandas":
+                yield batch.to_pandas(
+                    use_threads=True,
+                    split_blocks=True,
+                    self_destruct=True,
+                    types_mapper=ArrowDtype,
+                )
+            case "polars":
+                yield pl.from_arrow(batch)
+            case "arrow":
+                yield batch
 
 
 def query(
     *selectors: list[Selector],
     resolution_name: str | None = None,
     combine_type: Literal["concat", "explode", "set_agg"] = "concat",
-    return_type: Literal["pandas", "arrow"] = "pandas",
+    return_type: ReturnTypeStr = "pandas",
     threshold: int | None = None,
     limit: int | None = None,
     batch_size: int | None = None,
     return_batches: bool = False,
-) -> DataFrame | pa.Table | Iterator[DataFrame] | Iterator[pa.Table]:
+) -> QueryReturnType | Iterator[QueryReturnType]:
     """Runs queries against the selected backend.
 
     Args:
@@ -231,7 +237,7 @@ def query(
 
     Returns:
         If return_batches is False:
-            Data in the requested return type (DataFrame or pa.Table)
+            Data in the requested return type (DataFrame or ArrowTable)
         If return_batches is True:
             An iterator yielding batches in the requested return type
 
@@ -264,7 +270,7 @@ def query(
     if combine_type not in ("concat", "explode", "set_agg"):
         raise ValueError(f"combine_type of {combine_type} not valid")
 
-    if return_type not in ("pandas", "arrow"):
+    if return_type not in get_args(ReturnTypeStr):
         raise ValueError(f"return_type of {return_type} not valid")
 
     if not selectors:
@@ -342,15 +348,18 @@ def query(
                 result = result.rename_columns(rename_rule)
 
         # Return in requested format
-        if return_type == "pandas":
-            return result.to_pandas(
-                use_threads=True,
-                split_blocks=True,
-                self_destruct=True,
-                types_mapper=ArrowDtype,
-            )
-
-        return result
+        match return_type:
+            case "pandas":
+                return result.to_pandas(
+                    use_threads=True,
+                    split_blocks=True,
+                    self_destruct=True,
+                    types_mapper=ArrowDtype,
+                )
+            case "polars":
+                return pl.from_arrow(result)
+            case "arrow":
+                return result
 
 
 def match(
--- a/src/matchbox/common/db.py
+++ b/src/matchbox/common/db.py
@@ -20,78 +20,26 @@ T = TypeVar("T")
 def sql_to_df(
     stmt: Select,
     engine: Engine,
-    return_type: Literal["arrow"],
+    return_type: Literal["arrow", "pandas", "polars"],
     *,
     iter_batches: Literal[False] = False,
     batch_size: int | None = None,
     schema_overrides: dict[str, Any] | None = None,
     execute_options: dict[str, Any] | None = None,
-) -> ArrowTable: ...
+) -> QueryReturnType: ...
 
 
 @overload
 def sql_to_df(
     stmt: Select,
     engine: Engine,
-    return_type: Literal["arrow"],
+    return_type: Literal["arrow", "pandas", "polars"],
     *,
     iter_batches: Literal[True],
     batch_size: int | None = None,
     schema_overrides: dict[str, Any] | None = None,
     execute_options: dict[str, Any] | None = None,
-) -> Iterator[ArrowTable]: ...
-
-
-@overload
-def sql_to_df(
-    stmt: Select,
-    engine: Engine,
-    return_type: Literal["pandas"],
-    *,
-    iter_batches: Literal[False] = False,
-    batch_size: int | None = None,
-    schema_overrides: dict[str, Any] | None = None,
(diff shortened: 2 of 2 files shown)


src/matchbox/client/helpers/index.py
```diff
@@ -27,6 +27,7 @@ def index(
     engine: Engine,
     resolution_name: str | None = None,
     columns: list[str] | list[dict[str, dict[str, str]]] | None = None,
+    batch_size: int | None = None,
```
leo-mazzone, line 30: why are we exposing this in front-facing functions, instead of what we were doing i.e. it being configurable as a global option?

src/matchbox/common/db.py
```diff
@@ -34,58 +30,187 @@ def _convert_large_binary_to_binary(table: pa.Table) -> pa.Table:
...
+    schema_overrides: dict[str, Any] | None = None,
+    execute_options: dict[str, Any] | None = None,
+) -> (
+    pa.Table
+    | DataFrame
+    | PolarsDataFrame
+    | Iterator[pa.Table]
+    | Iterator[DataFrame]
+    | Iterator[PolarsDataFrame]
+):
```
leo-mazzone, line 125: It feels like 6 overloaded signatures are a bit much. Could we instead do
```
... -> QueryReturn | Iterator[QueryReturn]
```
and define `QueryReturn` as the union of all possible return types?

src/matchbox/common/db.py
```diff
@@ -34,58 +30,187 @@ def _convert_large_binary_to_binary(table: pa.Table) -> pa.Table:
...
+
+        if return_type == "polars":
+            return pl_result
+        elif return_type == "arrow":
+            arrow_table = pl_result.to_arrow()
+            return _convert_large_binary_to_binary(table=arrow_table)
+        elif return_type == "pandas":
+            return pl_result.to_pandas()
+        else:
+            raise ValueError(f"Unsupported return_type: {return_type}")
```
leo-mazzone, line 213: I've started checking for these things at the beginning to avoid useless waiting with a wrong argument. Worth doing it here?

src/matchbox/common/db.py
```diff
@@ -34,58 +30,187 @@ def _convert_large_binary_to_binary(table: pa.Table) -> pa.Table:
...
+        # Use the most efficient method for a single result
+        # Fall back if the URI method fails
+        try:
+            pl_result = pl.read_database_uri(
+                query=sql_query,
+                uri=url,
+                schema_overrides=schema_overrides,
+                execute_options=execute_options,
+            )
+        except Exception:
```
leo-mazzone, line 196: which Exception?

src/matchbox/common/db.py
leo-mazzone: It should also be consolidated in the main IF above

src/matchbox/common/db.py
```diff
@@ -34,58 +30,187 @@ def _convert_large_binary_to_binary(table: pa.Table) -> pa.Table:
...
     if not isinstance(url, str):
         raise ValueError("Unable to obtain a valid connection string from the engine.")
 
-    result = cx.read_sql(conn=url, query=sql_query, return_type=return_type)
-
-    if return_type == "arrow":
-        return _convert_large_binary_to_binary(table=result)
-
-    return result
+    if iter_batches:
```
leo-mazzone, line 165: Feels like this if/else could be removed and the conditional pushed down

test/client/test_helpers.py
```diff
@@ -693,6 +771,61 @@ def test_index_upload_failure(
...
+    source = source_factory(
+        features=[{"name": "company_name", "base_generator": "company"}],
+        engine=sqlite_warehouse,
+    )
+    mock_source_instance = source.mock
+    # Mock hash_data to capture the batch_size parameter
+    mock_source_instance.hash_data.return_value = pa.table(
+        {
+            "source_pk": [["1", "2"]],
+            "hash": pa.array([b"hash1"], type=pa.binary()),
```
leo-mazzone, line 789: not large_binary?

src/matchbox/client/helpers/selector.py
```diff
@@ -174,70 +275,87 @@ def query(
...
+            return result
+        elif return_type == "pandas":
+            return result.to_pandas(
+                use_threads=True,
+                split_blocks=True,
+                self_destruct=True,
+                types_mapper=ArrowDtype,
+            )
+        else:
+            raise ValueError(f"return_type of {return_type} not valid")
```
leo-mazzone, line 358: already checked above

src/matchbox/client/helpers/selector.py
```diff
@@ -94,26 +94,108 @@ def select(
...
+            yield batch
+        elif return_type == "pandas":
+            yield batch.to_pandas(
+                use_threads=True,
+                split_blocks=True,
+                self_destruct=True,
+                types_mapper=ArrowDtype,
+            )
+        else:
+            raise ValueError(f"return_type of {return_type} not valid")
```
leo-mazzone, line 176: again, should be handled elsewhere

src/matchbox/client/helpers/selector.py
```diff
@@ -94,26 +94,108 @@ def select(
...
+                if selector.fields:
+                    keep_cols = ["id"] + [
+                        selector.source.format_column(f) for f in selector.fields
+                    ]
+                    match_cols = [
+                        col for col in joined_table.column_names if col in keep_cols
+                    ]
+                    yield joined_table.select(match_cols)
+                else:
+                    yield joined_table
```
leo-mazzone, line 153: it doesn't feel good you're repeating this in a very similar way in the iterator version of the query function

src/matchbox/client/helpers/selector.py
```diff
@@ -94,26 +94,108 @@ def select(
...
+
+        selector_iters.append(process_batches(raw_batches, selector, mb_ids))
```
leo-mazzone, line 155: is this actually lazy?

src/matchbox/common/db.py
leo-mazzone: but do we really need the return type matrix?

src/matchbox/common/db.py
```diff
@@ -1,91 +1,187 @@
...
+
+        match return_type:
+            case "polars":
+                return results
+            case "arrow":
+                return (batch.to_arrow() for batch in results)
+            case "pandas":
+                return (batch.to_pandas() for batch in results)
+            case _:
+                raise ValueError(f"Unsupported return_type: {return_type}")
```
leo-mazzone, line 167: if we're confident in our logic, this line is impossible to reach. Remove?

src/matchbox/client/helpers/selector.py
```diff
@@ -94,6 +94,36 @@ def select(
...
+    # Apply field filtering if needed
+    if selector.fields:
+        keep_cols = ["id"] + [selector.source.format_column(f) for f in selector.fields]
+        match_cols = [col for col in joined_table.column_names if col in keep_cols]
+        return joined_table.select(match_cols)
+    else:
+        return joined_table
+
+
 def _query_batched(
```
leo-mazzone, line 127: this one is also not very DRY. The only part that differs from the loop, query, source.to_arrow  is the very bottom of the function. Could you instead do
```
for selector, sub_limit in zip(...):
    ...
   if batch_mode:
       yield stuff
    else:
        concatenate stuff
```

src/matchbox/client/helpers/index.py
leo-mazzone: Let's discuss this

src/matchbox/common/db.py
leo-mazzone: and same in the other match statement

leo-mazzone (approved):

PR #98, opened 2025-03-25
Add client version to request headers

ghost:
# Context

We may need to handle different versions of the Matchbox client. This PR adds client information to the request header so we can use it on the server side, should we need to.

Out of scope is doing anything on the server side.

* Adds `X-Matchbox-Client-Version` custom header

## Guidance to review

Nothing special.

## Relevant links

* [Custom headers in HTTPX](https://www.python-httpx.org/quickstart/#custom-headers)
* [Custom headers in FastAPI](https://fastapi.tiangolo.com/tutorial/header-params/)

## Checklist:

### DIFF
--- a/src/matchbox/client/_handler.py
+++ b/src/matchbox/client/_handler.py
@@ -2,6 +2,7 @@
 
 import time
 from collections.abc import Iterable
+from importlib.metadata import version
 from io import BytesIO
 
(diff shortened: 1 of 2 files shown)


sophie-daintta (approved):
LGTM

PR #99, opened 2025-03-25
Prevented model endpoints from accessing non-model Resolutions

ghost:
# Context

Previously, you could send strings for `Resolutions` that referred to sources and the client would error trying to make a `Model` object of them. Now the API just refuses.

Minor filter update to the common `resolve_model_name()` utility function.

## Guidance to review

Nothing special.

## Relevant links

None.

## Checklist:


leo-mazzone (approved):
Short, concise, elegant

PR #101, opened 2025-03-26
Single hashing space for clusters

leo-mazzone:
# Context

* Removed column aliases
* Removed signature property of sources
* Instead, hashes of source rows also include the sorted column names
* We can attempt to insert source clusters with the same hash and a different resolution

## Guidance to review

* I'm slightly unsure about is how expensive it will get to check new clusters against all existing ones. I suppose this part of the code will be changed down the line if we implement better upsert strategies.
* I've introduced the first emoji in the codebase (in warehouse.py). Am I bad? I was looking for a character that would be very unlikely to be found in a table. But maybe the better solution is to concatenate into a list rather than into a string.

## Relevant links

None so far.

## Checklist:


leo-mazzone:
The source signature was used in two ways:
- To prefix cluster hashes. This is not a problem anymore, same cluster across datasets is now allowed. HMRC importers and exporters can share hashes
- To name a resolution. Now we're just using the source address without any info about the columns. This means that it won't be possible (for now) to add sources with the same address and different columns, without overwriting the previous source. Down the line, namespacing source configs will take care of this

PR #106, opened 2025-04-01
Add logging to handler

ghost:
# Context

It's hard to diagnose problems in the DAG because there's no client-side logging. This adds it.

It also tidies up logging overall.

* Adds `logger.debug()` level messages to the handler
* Removes `matchbox.common.logging`'s insistence on the "info" level of logging
* Stripped all the logging config so it's in line with [Python documentation's advice](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library)

> It is strongly advised that you do not add any handlers other than [NullHandler](https://docs.python.org/3/library/logging.handlers.html#logging.NullHandler) to your library’s loggers. This is because the configuration of handlers is the prerogative of the application developer who uses your library. 

* Adds minor pytest configuration to silence noisy logs

## Guidance to review

Run `test_helpers.py`. Observe: no logging.

In `pyproject.toml`, set `log_cli = true`. Run `test_helpers.py`. Observe: useful debug appears, controlled solely by the user and root logger.

## Relevant links

* [Python documentation of library logging](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library)

## Checklist:

### DIFF
--- a/src/matchbox/common/logging.py
+++ b/src/matchbox/common/logging.py
@@ -39,13 +39,22 @@ class PrefixedLoggerAdapter(logging.LoggerAdapter):
         return msg, kwargs
 
 
-logger: Final[logging.Logger] = logging.getLogger("matchbox")
+logger: Final[PrefixedLoggerAdapter]
 """Logger for Matchbox.
 
 Used for all logging in the Matchbox library.
+
+Allows passing a prefix to any logging call.
+
+Examples:
+    ```python
+    log_prefix = f"Model {name}"
+    logger.debug("Inserting metadata", prefix=log_prefix)
+    logger.debug("Inserting data", prefix=log_prefix)
+    logger.info("Insert successful", prefix=log_prefix)
+    ```
 """
-logger.addHandler(logging.NullHandler())
-logger = PrefixedLoggerAdapter(logger, {})
+logger = PrefixedLoggerAdapter(logging.getLogger("matchbox"), {})
 
 
 console: Final[Console] = Console()


pyproject.toml
```diff
@@ -116,7 +116,7 @@ pythonpath = ["."]
 addopts = "--cov=matchbox test/ --log-disable=pg_bulk_ingest"
 norecursedirs = "src"
 log_cli = false  # Set to true to enable logs and individual tests
-log_cli_level = "ERROR"  # Set to "INFO" to see more detailed logging
+log_cli_level = "DEBUG"  # Set to "INFO" to see less detailed logging
```
leo-mazzone, line 119: Should "noisy" be the default when one clones the project?

src/matchbox/client/_handler.py
```diff
@@ -121,6 +122,9 @@ def query(
     threshold: int | None = None,
     limit: int | None = None,
 ) -> BytesIO:
+    log_prefix = f"[Query {str(source_address)[:10]}]"
```
leo-mazzone, line 125: the first 10 characters?

leo-mazzone: Gosh, I would get the whole full_name and the first 3 characters after the "at" followed by "...". Doing that abstraction I mention in the other comment will help

src/matchbox/client/_handler.py
```diff
@@ -175,6 +187,8 @@ def match(
         ),
     )
 
+    logger.debug(f"{log_prefix} Finished")
```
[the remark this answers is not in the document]
leo-mazzone: Yeah, so instead of doing it properly for now, could you just standardise the way you generate these statements in this file with two 1-line functions? e.g. `shorten_address()`, `add_log_prefix(prefix, message)`?

PR #108, opened 2025-04-02
Add batching to DAGs

ghost:
# Context

DAGs can't batch, which knocks over warehouses in pipelines that use them.

This adds the ability to batch to `StepInput` and `DAG.add_source()`.

Bonus 1: I finally clocked we were missing `Source.to_polars()` so I added it. We basically allow Arrow/Polars/Pandas everywhere except there, and as we're moving the codebase to Polars, this was a weird oversight.

Bonus 2: There was a huge hole in our docs -- `matchbox.common.arrow`, `.dtos`, `.sources` and `matchbox.client.dags`. I added them and fixed a few missing bits.

* `StepInput` gains a batch argument
* `DAG.add_source()` gains a batch argument
* Throughout the stack, better-handle the following three scenarios:
    * Don't batch, query in one
    * Batch, but concatenate it for me
    * Batch, and pass me the iterator
* Improve the unit tests for batching so they better-cover its usage (both index and query)
* Adds `Source.to_polars()` and unit tests conversion
* Adds documentation for `matchbox.common.arrow`, `.dtos` and `.sources`, and `matchbox.client.dags`

## Guidance to review

None.

## Relevant links

None.

## Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "matchbox-db"
-version = "0.0.0.dev0"
+dynamic = ["version"]
 description = "A framework for orchestrating and comparing data linking and deduplication methodologies."
 authors = [{ name="Department for Business and Trade" }]
 license = {file = "LICENSE"}
@@ -67,8 +67,12 @@ typing = [
 "Repository" = "https://github.com/uktrade/matchbox.git"
 
 [build-system]
-requires = ["setuptools>=42"]
+requires = ["setuptools>=64", "setuptools-scm>=8"]
 build-backend = "setuptools.build_meta"
+
+# Presence enables setuptools-scm
+[tool.setuptools_scm]
+
 
 [tool.uv]
 default-groups = ["dev", "typing"]
--- a/src/matchbox/server/Dockerfile
+++ b/src/matchbox/server/Dockerfile
@@ -25,6 +25,9 @@ COPY ./src/matchbox /code/src/matchbox
 # Then, add the rest of the project source code and install it
 # Installing separately from its dependencies allows optimal layer caching
 RUN --mount=type=cache,target=/root/.cache/uv \
+    # Determine package version dynamically from the git tag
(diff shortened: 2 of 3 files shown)


test/client/test_helpers.py
```diff
@@ -787,7 +787,6 @@ def test_index_with_batch_size(
 
     # Verify batch_size was passed to hash_data
     mock_source_instance.hash_data.assert_called_once()
-    assert mock_source_instance.hash_data.call_args.kwargs["iter_batches"] is True
     assert mock_source_instance.hash_data.call_args.kwargs["batch_size"] == 1000
```
leo-mazzone: Should there not be an equivalent line on the right?

src/matchbox/client/dags.py
```diff
@@ -177,6 +177,8 @@ def __init__(self):
         self.graph: dict[str, list[str]] = {}
         self.sequence: list[str] = []
 
+        self._index_batch_sizes: dict[str, int | None] = {}
```
leo-mazzone, line 180: Probably fine for the interim (?) but definitely weird that you specify batch size as an attribute to the step input for model sources, but it's a DAG attribute for indexing. Incidentally, any reason you might want to customise this on a source by source basis instead of making this global to the DAG?

src/matchbox/common/arrow.py
```diff
@@ -1,21 +1,29 @@
...
 
 from io import BytesIO
+from typing import Final
 
 import pyarrow as pa
 import pyarrow.parquet as pq
 
-SCHEMA_MB_IDS = pa.schema([("id", pa.int64()), ("source_pk", pa.large_string())])
-SCHEMA_INDEX = pa.schema(
+SCHEMA_MB_IDS: Final[pa.Schema] = pa.schema(
```
leo-mazzone, line 9: Don't mind the `Final` here, but what does it help us with? See this as a potential slippery slope as basically all our variables are meant to be final

leo-mazzone (approved):
Approving latest changes

PR #110, opened 2025-04-03
Change DB dependencies to allow async

leo-mazzone:
# Context

* Remove connectorx. Use ADBC instead
* Use the real polars, instead of polars-lts-cpu
* Upgrade psycopg2 to psycopg (but retain psycopg2 as a dev dependency for some tests, the ones that need to check different drivers not affecting the source address). Also install its async dependencies
* Install async extension for SQLAlchemy
* Make a few small changes to get tests to pass following changes in dependencies

## Guidance to review

I agree, the way I'm passing both a SQLAlchemy engine as well as an adbc_connection to sql_to_df is ugly, but it also... works? The thing is that we're using SQLAlchemy for preparing statements, even when we don't use it for running those statements.

There is an alternative to this that I've been toying with: continuing to use polars.read_database_uri, using the ADBC mode instead of the connectorx mode. It seems to be [suggested by the ADBC docs](https://arrow.apache.org/adbc/14/python/recipe/postgresql.html#using-pandas-and-adbc) as well. However, it negates batching, and I couldn't find proof one would be faster than another. Maybe we should do an empirical test (later?).

Another thing I'm not doing is connection pooling - see [here](https://arrow.apache.org/adbc/14/python/recipe/postgresql.html#connection-pooling-with-sqlalchemy). I am creating a new connection every time I need it, which probably will take about half a second. It feels like it will be comparatively much shorter than the query if calling sql_to_df.

## Relevant links

* [Installation instructions for psycopg](https://www.psycopg.org/psycopg3/docs/basic/install.html)
* From [ADBC docs](https://arrow.apache.org/adbc/0.3.0/python/quickstart.html): "the connection must be closed after usage or memory may leak"

## Checklist:


src/matchbox/server/postgresql/db.py
```diff
@@ -70,49 +93,125 @@ def get_engine(self) -> Engine:
...
 
-        self.engine.dispose()
+        # Dispose of all connections in the pool
+        if self.engine:
+            self.engine.dispose()
+
+        # Close the source ADBC connection if it exists
+        if self.adbc_connection:
+            self.adbc_connection.close()
+            self.adbc_connection = None
```
leo-mazzone, line 195: repeated just a few lines below

src/matchbox/server/postgresql/db.py
```diff
@@ -70,49 +93,125 @@ def get_engine(self) -> Engine:
...
+        if self.adbc_connection is None:
+            self.setup_adbc_pool()
+
+        @contextmanager
+        def connection_context() -> Generator[adbc_dbapi.Connection, None, None]:
+            # Clone a new connection from the source
+            conn = (
+                self.adbc_connection.adbc_clone()
+                if self.adbc_connection
+                else adbc_dbapi.connect(self.connection_string(driver=False))
```
leo-mazzone, line 155: you don't need to create a new connection if you setup the pool above

src/matchbox/server/postgresql/db.py
```diff
@@ -70,49 +93,125 @@ def get_engine(self) -> Engine:
...
+            finally:
+                session.close()
+
+        return session_context()
+
+    def setup_adbc_pool(self) -> None:
+        """Initialize the ADBC source connection for cloning."""
+        if self.adbc_connection is None:
+            # Create a source connection that will be used to clone from
+            self.adbc_connection = adbc_dbapi.connect(
```
leo-mazzone, line 129: can you immediately close this connection and still use it to clone from?

src/matchbox/server/postgresql/db.py
```diff
@@ -70,49 +93,125 @@ def get_engine(self) -> Engine:
...
+        def session_context() -> Generator[Session, None, None]:
+            session = cast(sessionmaker, self.SessionLocal)()
+            try:
+                yield session
+            finally:
+                session.close()
+
+        return session_context()
+
+    def setup_adbc_pool(self) -> None:
```
leo-mazzone, line 125: what's the point of this? You're not actually using a connection pool for ADBC. You're just cloning from a previous connection and immediately opening the cloned connection, which is as expensive as what I was doing before

leo-mazzone: This https://arrow.apache.org/adbc/14/python/recipe/postgresql.html#using-pandas-and-adbc
is just passing the cloning function to the pool manager, and letting it initialise new clones, not calling the cloning function on the spot

src/matchbox/common/db.py
```diff
@@ -3,27 +3,96 @@
...
+        # SQLAlchemy compiles for SQLAlchemy
+        return str(compiled_stmt)
+    else:
+        # SQLGlot transpiles for ADBC
+        target_dialect = detect_dialect(connection=connection)
+        return sqlglot.transpile(
+            sql=str(compiled_stmt),
+            read="postgres",
+            write=target_dialect,
+        )[0]
```
leo-mazzone, line 87: Slightly unsure about the SQLGlot business. Was it really already in the stack? Was it through SQLAlchemy? It feels weird to me that if you have a dialect-agnostic SQLAlchemy statement (which we already do everywhere we use sql_to_df) and use ADBC (which we also now do everywhere), you first compile it to Postgres and then transpile it to your target, when you could have compiled it correctly in the first instance. Can you not instead do

```python
compiled_stmt = stmt.compile(
    dialect=detect_dialect(connection=connection),
    compile_kwargs={"literal_binds": True},
)
```

leo-mazzone:
Added a few comments as I'm thinking about these issues as part of my L&D today.

PR #111, opened 2025-04-04
Fix deletion problem

ghost:
# Context

Deletion was happening at the ORM layer and therefore taking forever. This fixes it.

```python
probabilities = relationship(
    "Probabilities",
    back_populates="proposes",
    cascade="all, delete-orphan",
    passive_deletes=True,  # This is the solution!
)
```

There's also some minor formatting of integers in strings and the default `echo=False` in the SQLAlchemy engine is made explicit, as it's the thing you change to `True` to see logging.

## Guidance to review

None.

## Relevant links

* [Cascades documentation from SQLAlchemy](https://docs.sqlalchemy.org/en/20/orm/cascades.html)

## Checklist:

### DIFF
--- a/src/matchbox/server/postgresql/db.py
+++ b/src/matchbox/server/postgresql/db.py
@@ -60,7 +60,7 @@ class MatchboxDatabase:
         """Connect to the database."""
         if not self.engine:
             self.engine = create_engine(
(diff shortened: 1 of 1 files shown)


leo-mazzone (approved):
LGTM

PR #115, opened 2025-04-07
Change query to filter warehouse data client-side

ghost:
# Context

We were overwhelming a small warehouse with PKs in `query()`. This PR removes the limit argument and does all filtering client-side.

It's all in`matchbox.client.helpers.selector`.

* Processing moved to Polars
* Removed all limit arguments
* Factors out the batch processing as @leo-mazzone requested in #95 but that I couldn't think how to do at the time

I also dealt with the below deprecation warning:

```
SADeprecationWarning: The _ConnectionFairy.connection attribute is deprecated; please use 'driver_connection' (deprecated since: 2.0)
```

## Guidance to review

Confirm you're happy with using Polars here.

## Relevant links

None.

## Checklist:

### DIFF
--- a/src/matchbox/client/helpers/selector.py
+++ b/src/matchbox/client/helpers/selector.py
@@ -113,11 +113,9 @@ def _process_query_result(
 
 def _source_query(
     selector: Selector,
-    mb_ids: PolarsDataFrame | None = None,
     return_batches: bool = False,
     batch_size: int | None = None,
     only_indexed: bool = False,
-    limit: int | None = None,
 ) -> tuple[Source, Iterator[PolarsDataFrame]]:
     """From a Selector, query a source and join to matchbox IDs."""
     source = _handler.get_source(selector.address).set_engine(selector.engine)
@@ -136,10 +134,8 @@ def _source_query(
 
     raw_results = source.to_polars(
         fields=selected_fields,
-        pks=mb_ids["source_pk"].to_list() if mb_ids is not None else None,
         return_batches=return_batches,
         batch_size=batch_size,
-        limit=limit,
     )
 
     if isinstance(raw_results, PolarsDataFrame):
@@ -150,7 +146,6 @@ def _source_query(
 
 def _process_selectors(
     selectors: list[Selector],
-    sub_limits: list[int] | list[None],
     resolution_name: str | None,
     threshold: int | None,
     batch_size: int | None,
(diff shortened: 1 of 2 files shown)


src/matchbox/client/helpers/selector.py
```diff
@@ -136,78 +134,133 @@ def _source_query(
...
+    if isinstance(raw_results, PolarsDataFrame):
+        raw_results = [raw_results]
+
     return source, raw_results
 
 
-def _query_batched(
+def _process_selectors(
     selectors: list[Selector],
     sub_limits: list[int | None],
```
leo-mazzone, line 153: I think this should actually be list[int] | list[None]

leo-mazzone (approved):
LGTM

src/matchbox/client/helpers/selector.py
```diff
@@ -136,78 +134,97 @@ def _source_query(
...
+    client_side_filtering: bool = False
+
+    if not isinstance(sub_limits[0], int):
+        # List of None: no limit specified
+        client_side_filtering = True
+    elif any(sub_limit > limit_threshold for sub_limit in sub_limits):
+        logger.warning(
+            f"Limit is above {limit_threshold:,} threshold for "
+            "filtering warehouse-side. Filtering client-side."
+        )
```
leo-mazzone, line 175: should we not be warning in both cases?

src/matchbox/client/helpers/selector.py
```diff
@@ -136,78 +134,97 @@ def _source_query(
...
+
+    For non-batched queries, turn this into a list.
 
-    Returns an iterator yielding batches of results.
+    For batched queries, yield from it.
     """
+    limit_threshold: int = 1_000
+    client_side_filtering: bool = False
+
+    if not isinstance(sub_limits[0], int):
```
leo-mazzone, line 168: would be slightly more explicit to do
```
if sub_limits[0] is None:
```

src/matchbox/client/helpers/selector.py
```diff
@@ -136,78 +134,97 @@ def _source_query(
...
-    """Helper function that implements batched query processing.
+) -> Iterator[PolarsDataFrame]:
+    """Helper function to process selectors and return an iterator of results.
+
+    For non-batched queries, turn this into a list.
 
-    Returns an iterator yielding batches of results.
+    For batched queries, yield from it.
     """
+    limit_threshold: int = 1_000
```
leo-mazzone, line 165: if this were settable via an env variable, and there were a settings default of 1_000 then we could parameterise some of the existing tests and make sure they work both below as well as above this threshold?

src/matchbox/client/helpers/selector.py
```diff
@@ -136,78 +134,97 @@ def _source_query(
...
+        raw_results = [raw_results]
+
     return source, raw_results
 
 
-def _query_batched(
+def _process_selectors(
     selectors: list[Selector],
-    sub_limits: list[int | None],
+    sub_limits: list[int] | list[None],
```
leo-mazzone, line 153: I think this is a reasonable implementation of client-side limits. However, it looks like the complexity of dealing with limits keeps increasing, and I think we're yet to use them for anything? I wonder if it's still worth it?

PR #118, opened 2025-04-08
Add extra logging to the DAG

ghost:
# Context

I'm seeing errors in prod that confuse me and I felt that a visual depiction of the DAG and its current status would help.

This adds an `IndexStep` that simplifies the running of the DAG and opens up metadata that can be used to plot its current status.

Closes #117 

* Adds `IndexStep`
* Adds `DAG.draw()` and factors out its common code with `DAG.prepare()`
* Adds a unit test for `DAG.draw()`

## Guidance to review

None.

## Relevant links

* #117 

## Checklist:


src/matchbox/client/dags.py
```diff
@@ -13,10 +14,38 @@
...
+from matchbox.common.logging import logger
 from matchbox.common.sources import Source
 
-DAGNode = Union["ModelStep", Source]
-"""Type of node in the DAG. Either a step or a source."""
+DAGNode = Union["IndexStep", "ModelStep"]
+"""Type of node in the DAG. Either an indexing or model step."""
+
+
+class IndexStep(BaseModel):
```
leo-mazzone, line 24: Can we make sure all steps implement these by sub-classing an abstract class?

src/matchbox/client/dags.py
```diff
@@ -237,22 +289,134 @@ def depth_first(node: str, sequence: list):
...
+
+        # Add status indicator if start_time is provided
+        if start_time is not None:
+            node = self.nodes.get(root_name)
+
+            if root_name in skipped:
+                status = "⏭️"
+            elif doing and root_name == doing:
+                status = "🔄"
+            elif node and node.last_run and node.last_run > start_time:
```
leo-mazzone, line 327: what does this check mean? How would the last run timestamp be in the future? More generally, I'm not sure I get why you only get a status if a start_time is provided

src/matchbox/client/dags.py
```diff
@@ -237,22 +289,134 @@ def depth_first(node: str, sequence: list):
...
+                    elif doing and child == doing:
+                        status = "🔄"
+                    elif (
+                        child_node
+                        and child_node.last_run
+                        and child_node.last_run > start_time
+                    ):
+                        status = "✅"
+                    else:
+                        status = "⏸️"
```
leo-mazzone, line 366: this is the same as above

test/client/test_dags.py
```diff
@@ -450,7 +495,139 @@ def test_dag_disconnected(sqlite_warehouse: Engine):
     bar = source_factory(full_name="bar", engine=sqlite_warehouse).source
 
     dag = DAG()
-    dag.add_sources(foo, bar)
+    _ = dag.add_sources(foo, bar)
```
leo-mazzone, line 498: ?

leo-mazzone: does ruff complain about you just ignoring the objects as in the original line?

src/matchbox/client/dags.py
leo-mazzone: From what I can see, what the `@runtime_checkable` decorator enables is that you can do `isinstance(foo, Protocol`), which you're not doing. Maybe you could do it for each step you add to the DAG?

Though what I really had in mind was something static, not runtime, for us to reason about what we need to implement if we e.g. make changes to ModelStep. At the moment there isn't anything in the class definition itself that would remind us of what we promised to the DAG methods we would provide.

Another option is stop using properties (which are nice, but at the end, just syntactic sugar), and have e.g. `step.name()` everywhere, which allows to use abstract classes. What do you think?

leo-mazzone: 🦆

leo-mazzone (approved):
I really like this PR

PR #122, opened 2025-04-11
Fix Source ID bug and add basic async

ghost:
# Context

`Sources.resolution_id` and `Sources.source_id` were previously one, and were often identical in our unit tests. This masked some yawning bugs when they became different.

The key change is:

```python
class Sources(CountMixin, MBDB.MatchboxBase):
    """Table of sources of data for Matchbox."""

    __tablename__ = "sources"

    # Columns
    source_id = Column(
        BIGINT,
        Sequence("sources_source_id_seq", start=1, increment=1), # this forces them to be async in the first instance
        autoincrement=True,
        primary_key=True,
    )
```

That tests pass with this change shows the bug is fixed.

We also need our long-blocking endpoints (insert, query, match) to not block.

This fixes both these things.

* Change `Sources` autoincrement to being at 1 and correct a `primaryjoin` argument
* Change a multitude of places in `matchbox.server.postgres.utils.query` to fix the ensuing errors
* Remove `async` from the match, query and upload/backend processes functions, forcing FastAPI to [use a threadpool](https://fastapi.tiangolo.com/async/#path-operation-functions) and handle our async for us

> When you declare a path operation function with normal def instead of async def, it is run in an external threadpool that is then awaited, instead of being called directly (as it would block the server).

## Guidance to review

They key, key thing to be satisfied on is that my change to `Sources` will make the bug impossible to ignore in testing, and I can confirm it broke a million tests. If you're satisfied with this, and the tests now passing, you can review the changes to the utility functions safe in the knowledge that that basically work.

## Relevant links

* [FastAPI: concurrency and async / await](https://fastapi.tiangolo.com/async/)

## Checklist:

### DIFF
--- a/src/matchbox/server/postgresql/orm.py
+++ b/src/matchbox/server/postgresql/orm.py
@@ -7,8 +7,8 @@ from sqlalchemy import (
     CheckConstraint,
     Column,
     ForeignKey,
+    Identity,
     Index,
-    Sequence,
     UniqueConstraint,
     func,
     select,
@@ -179,7 +179,7 @@ class SourceColumns(CountMixin, MBDB.MatchboxBase):
     __tablename__ = "source_columns"
 
     # Columns
-    column_id = Column(BIGINT, primary_key=True, autoincrement=True)
+    column_id = Column(BIGINT, primary_key=True)
     source_id = Column(
         BIGINT,
         ForeignKey("sources.source_id", ondelete="CASCADE"),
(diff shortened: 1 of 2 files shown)


src/matchbox/server/postgresql/orm.py
```diff
@@ -240,7 +241,12 @@ class Sources(CountMixin, MBDB.MatchboxBase):
     __tablename__ = "sources"
 
     # Columns
-    source_id = Column(BIGINT, autoincrement=True, primary_key=True)
+    source_id = Column(
+        BIGINT,
+        Sequence("sources_source_id_seq", start=1, increment=1),
+        autoincrement=True,
```
sophie-daintta, line 247: I looking for an example of both Sequence and autoincrement being used but am having issues finding one. Just wanted to check they're both required?

src/matchbox/server/postgresql/utils/query.py
```diff
@@ -468,21 +490,21 @@ def _build_hierarchy_down(
...
-) -> list[tuple[Resolutions, str]]:
+) -> list[tuple[Sources, str]]:
     """Get target resolutions with source address."""
-    target_resolutions = []
+    target_sources: list[tuple[Sources, str]] = []
     for t in targets:
-        target_resolution = _get_dataset_resolution(t, session)
-        target_resolutions.append((target_resolution, t))
+        target_source = _get_dataset_source(t, session)
+        target_sources.append((target_source, t))
```
sophie-daintta, line 500: Just because a for loop was turned into a list comprehension elsewhere in the PR, would it make sense here?

sophie-daintta (approved):
Thanks, yep, the Identity update looks good. LGTM.

PR #123, opened 2025-04-11
Adds brute force migrations: drop the database if schema has changed

ghost:
# Context

Alembic is on the way: #112.

In the meantime, we can't deploy to staging or prod because schema changes in the database will mean we need DIAPP's help to drop the database.

So let's just drop the database.

* Adds `MatchboxDatabase.sync_schema()`, which drops and recreates the schema if there's any changes
* Calls it in `MatchboxPostgres.__init__()`, running it whenever the backend is initialised

## Guidance to review

This is a destructive update -- dangerous in a mature project.

This is a stopgap while we make proper migrations so we can get in front of users.

## Relevant links

* [Alembic's documentation on generating diffs](https://alembic.sqlalchemy.org/en/latest/api/autogenerate.html#alembic.autogenerate.compare_metadata)

## Checklist:

### DIFF
--- a/src/matchbox/server/postgresql/db.py
+++ b/src/matchbox/server/postgresql/db.py
@@ -171,11 +171,18 @@ class MatchboxDatabase:
                 return
 
         # Compare schema with ORM, drop and recreate if different
+        def _include_name(name: str, type_: str, _: dict[str, str]) -> bool:
+            if type_ == "schema":
+                return name == self.settings.postgres.db_schema
+            else:
+                return True
+
         with engine.connect() as conn:
             opts = {
                 "compare_type": True,
                 "compare_server_default": True,
                 "include_schemas": True,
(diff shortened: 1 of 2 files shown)


src/matchbox/server/postgresql/db.py
```diff
@@ -153,6 +156,38 @@ def clear_database(self):
...
+            if self.settings.postgres.db_schema not in schemas:
+                self.create_database()
+                return
+
+        # Compare schema with ORM, drop and recreate if different
+        with engine.connect() as conn:
+            opts = {
+                "compare_type": True,
+                "compare_server_default": True,
+                "include_schemas": True,
```
sophie-daintta, line 178: Happy to drop this one if you think its low risk, but maybe we could filter to just the mb schema?

sophie-daintta (approved):

PR #127, opened 2025-04-14
Removed _get_resolution_related_clusters as it no longer works

ghost:
# Context

Models couldn't be inserted because the `_get_resolution_related_clusters()` lookup took so long.

Refactoring isn't enough: allowing convergent clusters for different sources means the entire "family lookup" pattern no longer works.

As a quick and dirty fix, we now download the whole `Clusters` table. I believe this will work fine in the short term.

* Refactors database test scenarios to be easier to extend, and adds the `convergent` scenario to flush out the error I was getting with model ingestion
* Refactors testkits to hash the actual data they generate, better-reflecting the actual production environment
* 🐛 Removes `_get_resolution_related_clusters()` and downloads entire `Clusters` table on model results ingestion
* 🐛 `large_ingest()` was throwing deprecation warnings for `sa.Column.copy()` -- fixed
* 🐛 Adding truth-setting to end to end tests threw up a slew of places we weren't using it right (float to users, int to backend)
* 🐛 Deduplicates hierarchies when processing. I'll be honest: I don't know why there were duplicates but we just need it fixed right now

## Guidance to review

None.

## AI declaration

* Heavy use in refactoring scenarios. Code read and edited carefully, diffed with existing code it refactors

## Relevant links

None.

## Checklist:

### DIFF
--- a/test/fixtures/db.py
+++ b/test/fixtures/db.py
@@ -53,26 +53,19 @@ def register_scenario(name: str) -> Callable[[ScenarioBuilder], ScenarioBuilder]
 
 def _generate_cache_key(
     backend: MatchboxDBAdapter,
-    scenario_type: Literal["bare", "index", "dedupe", "link"],
+    scenario_type: str,
     warehouse: Engine,
     n_entities: int = 10,
     seed: int = 42,
-    **kwargs: dict[str, Any],
 ) -> str:
     """Generate a unique hash based on input parameters"""
-    extra_params = (
-        "_".join(f"{k}_{v}" for k, v in sorted(kwargs.items())) if kwargs else ""
-    )
+    if scenario_type not in SCENARIO_REGISTRY:
+        raise ValueError(f"Unknown scenario type: {scenario_type}")
 
-    cache_key = (
+    return (
         f"{warehouse.url}_{backend.__class__.__name__}_"
         f"{scenario_type}_{n_entities}_{seed}"
     )
-
-    if extra_params:
-        cache_key += f"_{extra_params}"
-
-    return cache_key
 
 
 def _testkitdag_to_warehouse(warehouse_engine: Engine, dag: TestkitDAG) -> None:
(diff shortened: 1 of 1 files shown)


leo-mazzone:
Is there a ticket to capture this?
>  I'll be honest: I don't know why there were duplicates but we just need it fixed right now

test/e2e/test_e2e.py
```diff
@@ -255,6 +255,7 @@ def _clean_company_name(df: DataFrame, prefix: str) -> DataFrame:
             assert identical, f"Deduplication of {source_name} failed: {report}"
 
             results.to_matchbox()
+            deduper.truth = 1.0
```
leo-mazzone, line 258: Is the idea that this is still a float because it's on the client side?

test/fixtures/db.py
```diff
@@ -35,18 +36,42 @@
...
     scenario_type: Literal["bare", "index", "dedupe", "link"],
     warehouse: Engine,
     n_entities: int = 10,
     seed: int = 42,
+    **kwargs: dict[str, Any],
 ) -> str:
     """Generate a unique hash based on input parameters"""
+    extra_params = (
+        "_".join(f"{k}_{v}" for k, v in sorted(kwargs.items())) if kwargs else ""
+    )
```
leo-mazzone, line 65: For joining, could you use a character that can't be used to name an argument? I.e. non-ASCII I guess?

test/fixtures/db.py
```diff
@@ -239,20 +297,104 @@ def create_scenario_dag(
...
+    **kwargs: dict[str, Any],
 ) -> Generator[TestkitDAG, None, None]:
     """Context manager for creating TestkitDAG scenarios."""
+    if scenario_type not in SCENARIO_REGISTRY:
+        raise ValueError(f"Unknown scenario type: {scenario_type}")
+
     # Generate cache key for backend snapshot
-    cache_key = _generate_cache_key(backend, scenario_type, warehouse, n_entities, seed)
+    cache_key = _generate_cache_key(
+        backend, scenario_type, warehouse, n_entities, seed, **kwargs
```
leo-mazzone, line 396: where are these kwargs actually used?

leo-mazzone (approved):
LGTM

PR #148, opened 2025-04-30
Add Location classes

ghost:
Locations will eventually replace `SourceAddress`. Adds them without hooking into `Source` yet.

The objects currently sit side by side, with tests for both.

* Adds `Location` ABC and first `RelationalDBLocation`
* Adds unit tests for each
* Adds SQL validation function and a new exception for it

## 👀 Guidance to review

I've deliberately not generalised these tests: new tests need writing for each location. I tried generalising and it was going to get very messy very quickly.

Try to think of the tests are aiming at the same coverage as `SourceAddress`.

In the end I didn't use [discriminated unions](https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions-with-str-discriminators), but rather a class constructor method on the Location ABC. Same same but different.

## 🤖 AI declaration

Helped with the unit tests but I rewrote most of them. Everything read and reviewed.

Significant AI input on the credentials test, which was a merge of two or three previous tests. Reviewed and edited.

## 🔗 Relevant links

* [Collections](https://uktrade.atlassian.net/wiki/spaces/CDL/pages/4788781057/Collections) ideation document

## ✅ Checklist:

### DIFF
--- a/src/matchbox/common/db.py
+++ b/src/matchbox/common/db.py
@@ -182,33 +182,20 @@ def validate_sql_for_data_extraction(sql: str) -> bool:
 
     expr = expressions[0]
 
-    if not isinstance(expr, (sqlglot.expressions.Select, sqlglot.expressions.With)):
+    if not isinstance(expr, sqlglot.expressions.Select):
         raise MatchboxSourceExtractTransformError(
             "SQL statement must start with a SELECT or WITH command."
         )
 
-    forbidden_classes = (
-        sqlglot.expressions.Insert,
-        sqlglot.expressions.Update,
-        sqlglot.expressions.Delete,
-        sqlglot.expressions.Create,
-        sqlglot.expressions.Drop,
-        sqlglot.expressions.Alter,
-        sqlglot.expressions.Grant,
-        sqlglot.expressions.TruncateTable,
+    forbidden = (
+        sqlglot.expressions.DDL,
+        sqlglot.expressions.DML,
+        sqlglot.expressions.Into,
     )
 
-    nodes = list(expr.find_all(forbidden_classes))
-    if nodes:
+    if len(list(expr.find_all(forbidden))) > 0:
         raise MatchboxSourceExtractTransformError(
-            f"SQL statement contains forbidden command: {nodes}"
+            "SQL statement must not contain DDL or DML commands."
         )
-
-    select_nodes = expr.find_all(sqlglot.expressions.Select)
-    for select in select_nodes:
-        if select.args.get("into"):
-            raise MatchboxSourceExtractTransformError(
-                "SQL statement contains a SELECT INTO command, which modifies data."
-            )
 
     return True
--- a/src/matchbox/common/sources.py
+++ b/src/matchbox/common/sources.py
@@ -174,9 +174,8 @@ class RelationalDBLocation(Location):
             raise ValueError("Credentials should not be in the URI.")
         if value.query or value.fragment:
             raise ValueError("Query params and fragments should not be in the URI.")
-        if value.scheme:
-            if "+" in value.scheme:
-                raise ValueError("Driver should not be in the URI.")
+        if "+" in value.scheme:
+            raise ValueError("Driver should not be in the URI.")
         return value
 
     def _validate_engine(self, credentials: Engine) -> None:
@@ -217,6 +216,10 @@ class RelationalDBLocation(Location):
             return False
 
     def validate_extract_transform(self, extract_transform: str) -> bool:  # noqa: D102
+        # We are NOT attempting a full sanitisation of the SQL statement
+        # Validation is done purely to stop accidental mistakes, not malicious actors
+        # Users should only run indexing using Sources they trust and have read,
+        # using least privilege credentials
         return validate_sql_for_data_extraction(extract_transform)
 
     @requires_credentials
--- a/test/common/test_sources.py
+++ b/test/common/test_sources.py
@@ -221,6 +221,19 @@ def test_relational_db_add_credentials(sqlite_warehouse: Engine):
             "SELECT * FROM users; DROP TABLE users;", False, id="multiple-statements"
         ),
         pytest.param("SELECT * INTO new_table FROM users", False, id="select-into"),
+        pytest.param(
+            """
+            WITH updated_rows AS (
+                UPDATE employees
+                SET salary = salary * 1.1
+                WHERE department = 'Sales'
+                RETURNING *
+            )
+            SELECT * FROM updated_rows;
+            """,
+            False,
+            id="non-query-cte",
+        ),
     ],
 )
 def test_relational_db_extract_transform(sql: str, is_valid: bool):


src/matchbox/common/sources.py
```diff
@@ -48,6 +61,168 @@
...
+    credentials: Any | None = Field(exclude=True, default=None)
+
+    @abstractmethod
+    def add_credentials(self, credentials: Any) -> None:
+        """Adds credentials to the location."""
+        ...
+
+    @abstractmethod
+    def connect(self) -> bool:
+        """Establish connection to the data source.
```
leo-mazzone, line 104: connection to the location?

src/matchbox/common/sources.py
```diff
@@ -48,6 +61,168 @@
...
+        """Execute ET logic against this location and return Arrow RecordBatches.
+
+        Raises:
+            AttributeError: If the credentials are not set.
+        """
+        ...
+
+    @classmethod
+    def create(cls, data: dict) -> "Location":
+        """Factory method to create the appropriate Location subclass."""
```
leo-mazzone, line 129: example usage please? Don't understand what would go in the dict

leo-mazzone: also when would you use this instead of calling RightClass.from_credentials?

src/matchbox/common/sources.py
```diff
@@ -48,6 +61,168 @@
...
+    type: Literal["dbms"] = "dbms"
+    credentials: Engine | None = Field(
+        exclude=True,
+        default=None,
+        description=(
+            "The credentials for a relational database are a SQLAlchemy Engine."
+        ),
+    )
+
+    @field_validator("uri", mode="after")
```
leo-mazzone, line 152: where's uri defined as an attribute?

src/matchbox/common/sources.py
```diff
@@ -48,6 +61,168 @@
...
+    def add_credentials(self, credentials: Engine) -> None:  # noqa: D102
+        self._validate_engine(credentials)
+        self.credentials = credentials
+
+    @requires_credentials
+    def connect(self) -> bool:  # noqa: D102
+        try:
+            _ = sql_to_df(stmt="select 1", connection=self.credentials)
+            return True
+        except Exception as _:
```
leo-mazzone, line 191: I keep being confused about all this underscores. Are they mandated by ruff?

src/matchbox/common/sources.py
```diff
@@ -48,6 +61,168 @@
...
+            return False
+
+    def validate_extract_transform(self, extract_transform: str) -> bool:  # noqa: D102
+        try:
+            sqlglot.parse_one(extract_transform)
```
leo-mazzone, line 196: Not immediately clear to me - will this reject something outside DQL?

test/common/test_sources.py
```diff
@@ -2,16 +2,246 @@
...
+        ),
+        pytest.param(
+            "sqlite:///test.db", "sqlite:///other.db", False, id="different-path"
+        ),
+        pytest.param(
+            "postgresql://localhost:5432/testdb",
+            "postgresql://otherhost:5432/testdb",
+            False,
+            id="different-host",
+        ),
```
leo-mazzone, line 76: being quite pedantic here, but I think the existing tests are much more comprehensive, and less verbose because they're using the unique-ing ability of sets instead of pairs in pytest.mark.parametrize

test/common/test_sources.py
```diff
@@ -2,16 +2,246 @@
...
+
+    # Verify the result structure
+    assert set(df.columns) == {"name", "company_name", "crn"}
+
+    # Verify the calculated columns
+    sample_str: str = df.select("company_name").row(0)[0]
+    assert sample_str == sample_str.upper()
+
+
+def test_handling_of_invalid_query(sqlite_warehouse: Engine):
```
leo-mazzone, line 220: could be made infinitesimely faster by using sqlite://:memory: instead of the fixture :D

test/common/test_sources.py
```diff
@@ -2,16 +2,246 @@
...
+
+@pytest.mark.parametrize(
+    ["sql", "is_valid"],
+    [
+        pytest.param("SELECT * FROM test_table", True, id="valid-select"),
+        pytest.param(
+            "SELECT id, name FROM test_table WHERE id > 1", True, id="valid-where"
+        ),
+        pytest.param("SLECT * FROM test_table", False, id="invalid-syntax"),
+        pytest.param("", False, id="empty-string"),
```
leo-mazzone, line 121: DROP TABLE important_table; ?

test/common/test_sources.py
```diff
@@ -2,16 +2,246 @@
...
+        # Should not raise an exception
+        location._validate_engine(engine)
+    else:
+        # Should raise ValueError
+        with pytest.raises(ValueError, match="does not match"):
+            location._validate_engine(engine)
+
+
+def test_credentials_connect(sqlite_warehouse: Engine):
+    """Test adding credentials and conncting to a RelationalDBLocation."""
```
leo-mazzone, line 94: connecting

src/matchbox/common/sources.py
leo-mazzone: It's also not establishing a connection in the sense of e.g. caching. I'd phrase it as "test connection"

test/common/test_sources.py
1lliter8: Doesn't work -- the engine doesn't have an URL.

src/matchbox/common/db.py
```diff
@@ -145,3 +148,67 @@ def fullname_to_prefix(fullname: str) -> str:
...
+        sqlglot.expressions.Insert,
+        sqlglot.expressions.Update,
+        sqlglot.expressions.Delete,
+        sqlglot.expressions.Create,
+        sqlglot.expressions.Drop,
+        sqlglot.expressions.Alter,
+        sqlglot.expressions.Grant,
+        sqlglot.expressions.TruncateTable,
+    )
+
```
leo-mazzone, line 200: I think the only reason this is needed is that we allow to start with a CTE. I realise it's nice to support, but I'm then concerned that forbidden classes will be a bit fragile. Having a quick look at sqlglot.expressions, I'm not super confident you've covered everything

src/matchbox/common/sources.py
```diff
@@ -174,6 +174,9 @@ def validate_uri(cls, value: AnyUrl) -> AnyUrl:
             raise ValueError("Credentials should not be in the URI.")
         if value.query or value.fragment:
             raise ValueError("Query params and fragments should not be in the URI.")
+        if value.scheme:
+            if "+" in value.scheme:
```
leo-mazzone, line 178: when is value.scheme True with no plus?

src/matchbox/common/db.py
```diff
@@ -182,33 +182,20 @@ def validate_sql_for_data_extraction(sql: str) -> bool:
 
     expr = expressions[0]
 
-    if not isinstance(expr, (sqlglot.expressions.Select, sqlglot.expressions.With)):
+    if not isinstance(expr, sqlglot.expressions.Select):
         raise MatchboxSourceExtractTransformError(
             "SQL statement must start with a SELECT or WITH command."
```
leo-mazzone, line 187: Update error message

PR #153, opened 2025-05-06
Implement evaluation features

leo-mazzone:
This is the bulk set of changes allowing users to sample data, send evaluation judgements, and for model builders to use these judgements to assess model precision and recall.

* Sketched beginnings of a user model - this does not do much yet, and has no "public" functions
* Added ability to sample clusters to evaluate, and to join to warehouse data
  * This means we now allow executing an extract/transform query at a location, while additionally filtering by key (filter pushdown)
* Added `EvalJudgements` table to the ORM, and auxiliary objects to form and retrieve judgements
  * Added associated operations across client, API and adapter
* Added functions that compute precision and recall client-side from model results and user judgements
  * This requires that the client can request leaf IDs as well as root IDs, which is now the default
    * This led to a generalisation of the unified query builder
  * This requires that one can now form root-leaf from client-side results
* Added backend feature to compare multiple models against validation data
* Added simple Streamlit application that operationalises sampling and forming / submitting judgements
  * Added wrapper to try out UI with mock data
* Tested evaluation features end-to-end

## 👀 Guidance to review
* This is a big PR, and you need to decide in which order you should look at different files. Perhaps a sensible approach would be to do a first pass where you start from the end to end test and look at what handler operations they call, then navigate to the corresponding API calls and auxiliary common objects, then the adapter, and finally the ORM.
* **Why am I sending zipped parquet tables?** The new pattern of zipping multiple files is motivated by these factors:
  * The server produces arrow tables, and arrow is also a convenient format for the client, because then we can have a common function usable by both sides that takes the exact same Arrow tables
  * The alternative, JSON, would require a translation back and forth, and would produce a much larger file to transfer. At the moment we're dealing with small data, but I believe this solution is both more convenient to implement as well as more future-proof
* **Why am I not averaging PR over users in the end?** It's not simple to make the argument either way. Averaging over users means that the impact of any user is capped, for better or for worse. For now, sticking to a simpler implementation, but it won't be difficult to average over users if we so desire.
* Please note that for the filter pushdown I deliberately chose to not use `sqlglot`
* **Why is downloading leaf IDs the default for the client's `query()`?** Because every time the user calls it directly, it's more likely they're building a model from scratch and want to evaluate it. Everywhere else the function is used (in the DAG, to extract the mappings table), it's abstracted away, and we can explicitly reject the leaf IDs and save some bandwidth
* **Where are the updates to guides in the docs?** They are pending the approval of this set of features.
* My changes to client `Results` made me spot that we're now using a special function to get clusters from probabilities on the client-side, even though it lives under `common`. I'd like to get to that at some point

## 🤖 AI declaration
* Used in minor ways to ask how to use libraries when I got too lazy to consult docs
* Used to produce short snippets (e.g. zipping unzipping) which have been adapted and verified carefully

## 🔗 Relevant links

* [JSON API specification](https://jsonapi.org/format/#crud-updating-responses-404): this inspired my approach to HTTP error codes, i.e. 404s vs 422s.

## ✅ Checklist:

- [x] All new code is tested
    - [ ] Tutorials
    - [ ] Developer docs


src/matchbox/client/eval/ui.py
```diff
@@ -0,0 +1,83 @@
...
+
+from matchbox.client import _handler
+from matchbox.client.eval import get_samples
+from matchbox.common.eval import Judgement
+from matchbox.common.graph import DEFAULT_RESOLUTION
+
+st.title("Matchbox evaluation session")
+
+if "step" not in st.session_state:
+    st.session_state.user_name = "scott.mcgregor"
```
will-langdale, line 14: Can we not hard code this in prod?

src/matchbox/client/eval/ui.py
```diff
@@ -0,0 +1,83 @@
...
+    st.session_state.user_id = _handler.login(user_name=st.session_state.user_name)
+    st.session_state.resolution = DEFAULT_RESOLUTION
```
will-langdale, line 16: This also feels like it need parameterising. Happy for it to be another ticket.

src/matchbox/client/eval/utils.py
```diff
@@ -0,0 +1,161 @@
...
+
+    return results_by_root
+
+
+class EvalData:
+    """Object which caches evaluation data to measure performance of models."""
+
+    def __init__(self):
+        """Initialise evaluation data from resolution name."""
+        self.judgements, self.expansion = _handler.download_eval_data()
```
will-langdale, line 110: For a future ticket, but downloading ALL eval data for all models seems like a bad idea.

test/common/test_eval.py
```diff
@@ -0,0 +1,204 @@
...
+            {"root": 5, "leaf": 5},
+            # (6,7)
+            {"root": 67, "leaf": 6},
+            {"root": 67, "leaf": 7},
+        ]
+    )
+
+    judgements = pa.Table.from_pylist(
+        [
+            # Ambiguoud but more positive than negative
```
will-langdale, line 88:
```suggestion
            # Ambiguous but more positive than negative
```

test/common/test_eval.py
```diff
@@ -0,0 +1,204 @@
...
+            {"root": 101, "leaves": [3, 4, 5]},
+            {"root": 102, "leaves": [6, 7]},
+            {"root": 103, "leaves": [8, 9]},
+            {"root": 104, "leaves": [10, 11]},
+            {"root": 105, "leaves": [12, 13]},
+            {"root": 106, "leaves": [12, 13, 14]},
+            # New "splintered" clusters
+            {"root": 200, "leaves": [3, 4]},
+        ]
+    )
```
will-langdale, line 176: Put this first, it's key to understanding the judgements.

will-langdale: This is the first time I've thought that `expansion` is the wrong word. These are just clusters, I think, and `clusters` before `judgements` makes sense in a way that `expansion` before `judgement` doesn't.

test/e2e/test_e2e_evaluation.py
```diff
@@ -0,0 +1,201 @@
...
+    SourceTestkitParameters,
+    SuffixRule,
+    linked_sources_factory,
+)
+from matchbox.common.sources import RelationalDBLocation, SourceConfig
+
+
+@pytest.mark.docker
+class TestE2EPipelineBuilder:
+    """End to end tests for DAG pipeline functionality."""
```
will-langdale, line 25:
```suggestion
class TestE2EModelEvaluation:
    """End to end tests for model evaluation functionality."""
```

test/fixtures/db.py
```diff
@@ -317,6 +317,94 @@ def create_link_scenario(
...
+
+        model_testkit2 = query_to_model_factory(
+            left_resolution=source.name,
+            left_query=source_query,
+            left_keys={source.name: "key"},
+            true_entities=tuple(linked.true_entities),
+            name=model_name2,
+            description=f"Deduplication of {source.name}",
+            prob_range=(0.5, 1.0),
+            seed=seed,
```
will-langdale, line 386: `seed + 1` so they have different probabilities and are marginally different models?

test/fixtures/db.py
```diff
@@ -317,6 +317,94 @@ def create_link_scenario(
...
+        )
+
+        assert model_testkit1.probabilities.num_rows > 0
+        assert model_testkit1.probabilities == model_testkit2.probabilities
+
+        for model, threshold in zip(
+            [model_testkit1, model_testkit2], [50, 75], strict=True
+        ):
```
will-langdale, line 394: Optional, but I personally find `zip` very unnerving and avoid it if I can. This relies less on us happening to line things up and is much clearer on what's being iterated over:

```suggestion
        for model, threshold in (
            (model_testkit1, 50),
            (model_testkit2, 75),
        ):
```

test/server/api/routes/test_routes_eval.py
```diff
@@ -0,0 +1,238 @@
...
+    app.dependency_overrides[backend] = lambda: mock_backend
+    judgement = Judgement(user_id=1, shown=10, endorsed=[[1]])
+    response = test_client.post("/eval/judgements", json=judgement.model_dump())
+    assert response.status_code == 201
+    assert (
+        mock_backend.insert_judgement.call_args_list[0].kwargs["judgement"] == judgement
+    )
+
+
+def test_insert_judgement_error(test_client: TestClient):
```
will-langdale, line 43: I think docstrings are generally useful and easy to add.

test/server/api/routes/test_routes_eval.py
```diff
@@ -0,0 +1,238 @@
...
+def test_get_judgements(test_client: TestClient):
+    """Test that all judgements can be retrieved."""
+    judgements = pa.Table.from_pylist(
+        [
+            {"user_id": 1, "shown": 10, "endorsed": 11},
+            {"user_id": 1, "shown": 10, "endorsed": 12},
+        ],
+        schema=SCHEMA_JUDGEMENTS,
+    )
+    # There will be nulls in case of a schema mismatch
```
will-langdale, line 71: Why don't we just make the schema non-nullable? There's also [a `.validate()` method](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.from_pylist) that might be more appropriate here.

test/server/test_adapter.py
```diff
@@ -901,3 +922,262 @@ def get_counts():
...
+            # Let's massage tables into a root-leaf dict for all endorsed clusters
+            endorsed_root, endorsed_leaves = (
+                pl.from_arrow(judgements)
+                .join(pl.from_arrow(expansion), left_on="endorsed", right_on="root")[
+                    ["endorsed", "leaves"]
+                ]
+                .to_dict(as_series=False)
+                .values()
+            )
+            endorsed_dict = dict(zip(endorsed_root, endorsed_leaves, strict=True))
```
will-langdale, line 1024: AI reckons this should work.

```suggestion
            endorsed_dict = dict(
                pl.from_arrow(judgements)
                .join(pl.from_arrow(expansion), left_on="endorsed", right_on="root")
                .select(["endorsed", "leaves"])
                .rows()
            )
```

src/matchbox/client/_handler.py
```diff
@@ -79,21 +94,34 @@ def handle_http_code(res: httpx.Response) -> httpx.Response:
...
         raise MatchboxDeletionNotConfirmed(message=error.details)
 
     if res.status_code == 422:
+        if (
+            "parameter" in res.json()
+            and res.json()["parameter"] == BackendParameterType.SAMPLE_SIZE
+        ):
+            raise MatchboxTooManySamplesRequested(res.content)
+        # Not a custom Matchbox exception, most likely a Pydantic error
         raise MatchboxUnparsedClientRequest(res.content)
```
will-langdale, line 120:
```suggestion
    match res.json().get("parameter"):
        case BackendParameterType.SAMPLE_SIZE:
            raise MatchboxTooManySamplesRequested(res.content)
        case _:
            # Not a custom Matchbox exception, most likely a Pydantic error
            raise MatchboxUnparsedClientRequest(res.content)
```

src/matchbox/client/results.py
```diff
@@ -112,6 +96,16 @@ def check_probabilities(cls, value: pa.Table | pl.DataFrame) -> pa.Table:
 
         return value.cast(SCHEMA_RESULTS)
 
+    @property
+    def clusters(self):
```
will-langdale, line 100: This is quite an expensive operation and I wonder if it should be cached. Perhaps `to_clusters` would be the place to do it?

src/matchbox/client/_handler.py
```diff
@@ -391,6 +433,58 @@ def delete_resolution(
...
+        with zip_file.open(JudgementsZipFilenames.EXPANSION) as f2:
+            expansion = read_table(f2)
+
+    logger.debug("Finished retrieving judgements.")
+
+    def check_schema(expected: Schema, actual: Schema) -> None:
+        if expected != actual:
+            raise MatchboxClientFileError(
+                message=(f"Schema mismatch. Expected:\n{expected}\nGot:\n{actual}")
+            )
```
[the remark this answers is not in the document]
leo-mazzone: Done

src/matchbox/client/eval/mock_ui.py
```diff
@@ -0,0 +1,114 @@
...
+    matchbox_client.delete("/database", params={"certain": "true"})
+
+
+if __name__ == "__main__":
+    atexit.register(cleanup_database)
+    warehouse_url = setup_mock_database()
+
+    environ["MB__CLIENT__DEFAULT_WAREHOUSE"] = str(warehouse_url)
+    subprocess.run(
+        [sys.executable, "-m", "streamlit", "run", "src/matchbox/client/eval/ui.py"]
```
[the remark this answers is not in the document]
leo-mazzone: we don't have any way of unit testing Streamlit either, so the only way is kind of playing with it ourselves

src/matchbox/common/eval.py
```diff
@@ -0,0 +1,227 @@
...
+    )
+
+    all_leaves = set(
+        chain.from_iterable(expanded_judgements["endorsed_leaves"].to_list())
+    )
+
+    if len(expanded_judgements) != len(judgements):
+        raise ValueError("Malformed judgements / expansion data received.")
+
+    validation_net_count: dict[Pair, int] = {}
```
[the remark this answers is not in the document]
leo-mazzone: Done.

src/matchbox/common/eval.py
```diff
@@ -0,0 +1,227 @@
...
+        strict=True,
+    ):
+        positive_pairs = set(combinations(sorted(endorsed), r=2))
+        potential_pairs = set(combinations(sorted(shown), r=2))
+
+        # For every pair implicitly rejected, we subtract from its count the proportion
+        # of the shown cluster endorsed in this row. This is to avoid double counting.
+        # For example, if user shown (123) and endorses (12) and (3), the pairs (13) and
+        # (23), which are rejected on two rows, will subtract -2/3 and -1/3 respectively
+        negative_pairs = potential_pairs - positive_pairs
```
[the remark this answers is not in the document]
leo-mazzone: Done.

src/matchbox/client/eval/utils.py
leo-mazzone: https://uktrade.atlassian.net/browse/MTCHBX-461?atlOrigin=eyJpIjoiZmQ4YzEwMTg2ZmJmNDRlZTg2YTRkMjExMmZmOWRjNmMiLCJwIjoiaiJ9

src/matchbox/common/sources.py
```diff
@@ -154,6 +155,9 @@ def execute(
                 * If a callable is provided, it will take the old name as input and
                     return the new name.
             return_type: The type of data to return. Defaults to "polars".
+            keys: Dictionary with a single key.
+                The key of the dictionary is a field name on which to filter.
+                Filters source entries where the key field is in the dict values.
```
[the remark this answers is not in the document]
leo-mazzone: Done.

src/matchbox/client/_handler.py
```diff
@@ -391,6 +433,58 @@ def delete_resolution(
...
+
+def send_eval_judgement(judgement: Judgement) -> None:
+    logger.debug(
+        f"Submitting judgement {judgement.shown}:{judgement.endorsed} "
+        f"for {judgement.user_id}"
+    )
+    CLIENT.post("/eval/judgements", json=judgement.model_dump())
+
+
+def download_eval_data() -> tuple[Table, Table]:
```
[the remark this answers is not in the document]
leo-mazzone: I have created that ticket and linked it to a different comment

test/server/test_adapter.py
```diff
@@ -901,3 +922,262 @@ def get_counts():
...
+            for shown, endorsed in zip(
+                root_leaves["root"].to_list(),
+                root_leaves["leaf"].to_list(),
+                strict=True,
+            ):
+                self.backend.insert_judgement(
+                    judgement=Judgement(
+                        user_id=user_id, shown=shown, endorsed=[endorsed]
+                    )
+                )
```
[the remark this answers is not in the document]
leo-mazzone: Done

src/matchbox/client/eval/utils.py
```diff
@@ -0,0 +1,161 @@
...
+
+        root_leaf = (
+            results.root_leaf()
+            .rename({"root_id": "root", "leaf_id": "leaf"})
+            .to_arrow()
+        )
+        return precision_recall([root_leaf], self.judgements, self.expansion)[0]
+
+    def pr_curve(self, results: Results) -> Figure:
+        """For each threshold in retults computes precision and recall."""
```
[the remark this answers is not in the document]
leo-mazzone: Done

test/server/postgresql/test_pg_sql.py
```diff
@@ -612,76 +612,86 @@ def test_get_existing_source(self, populated_postgres_db: MatchboxPostgres):
...
-@pytest.mark.parametrize("mode", ["root_leaf", "id_key"])
+@pytest.mark.parametrize(
+    ("level", "get_hashes"),
+    [
+        pytest.param("leaf", True),
+        pytest.param("leaf", False),
+        pytest.param("key", True),
+        pytest.param("key", False),
+    ],
+)
```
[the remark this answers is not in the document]
leo-mazzone: Done

src/matchbox/server/postgresql/orm.py
```diff
@@ -550,6 +551,42 @@ class Clusters(CountMixin, MBDB.MatchboxBase):
...
+    judgement_id = Column(BIGINT, primary_key=True)
+    user_id = Column(
+        BIGINT, ForeignKey("users.user_id", ondelete="CASCADE"), nullable=False
+    )
+    endorsed_cluster_id = Column(
+        BIGINT, ForeignKey("clusters.cluster_id", ondelete="CASCADE"), nullable=False
+    )
+    shown_cluster_id = Column(
+        BIGINT, ForeignKey("clusters.cluster_id", ondelete="CASCADE"), nullable=False
+    )
```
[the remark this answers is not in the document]
leo-mazzone: Deleting the model doesn't currently delete the cluster, however in the future we might decide orphaned clusters get deleted. But even then, a cluster with a judgement on it isn't orphaned, so I think it's safe to keep this cascade?

src/matchbox/server/postgresql/orm.py
```diff
@@ -550,6 +551,42 @@ class Clusters(CountMixin, MBDB.MatchboxBase):
...
+class EvalJudgements(CountMixin, MBDB.MatchboxBase):
+    """Table of evaluation judgements produced by human validators."""
+
+    __tablename__ = "eval_judgements"
+
+    # Columns
+    judgement_id = Column(BIGINT, primary_key=True)
+    user_id = Column(
+        BIGINT, ForeignKey("users.user_id", ondelete="CASCADE"), nullable=False
+    )
```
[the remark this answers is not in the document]
leo-mazzone: We don't want to keep a judgement without a user identity. In fact, I think that deleting a user would be exactly the mechanism we use to delete judgements. There is no other reason to delete a user I don't think - if a user leaves DBT or something, we can still store their user ID and user name

src/matchbox/common/eval.py
```diff
@@ -0,0 +1,227 @@
...
+    - If a validation pair was rejected more times than it was endorsed, remove it from
+        validation pairs, but keep it in model pairs.
+    - Precision and recall are computed for each model against validation pairs.
+
+    At the moment, this function ignores user IDs.
+
+    Args:
+        models_root_leaf: list of tables with root and leaf columns, one per model.
+            They must include all the clusters that resolve from a model, all the way
+            to the original source clusters if no model in the lineage merged them.
```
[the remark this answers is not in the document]
leo-mazzone: on the fence; the schemas have been particularly useful to track things moving between client and server. I think having a schema to describe every table that's input to a function in our code base would be overdoing it

src/matchbox/common/eval.py
```diff
@@ -0,0 +1,227 @@
...
+    vs. when the user is shown (123) and endorses (12). In the second case, the user
+    implies a negative judgement over pairs (1,3) and (2,3). We return the net value of
+    pairs by summing 1 for an endorsement and subtracting 1 for a rejection.
+
+    This function relies on the input data being well-formed. For example,
+
+    - All shown cluster IDs need to have an expansion.
+    - All endorsed cluster IDs need to have an expansion, unless they're leaves.
+    - No partial splintered clusters, i.e. if (123)->(12) is in the data, (123)->(3)
+        must be as well.
```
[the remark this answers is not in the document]
leo-mazzone: Can you make an example to illustrate?

src/matchbox/client/eval/mock_ui.py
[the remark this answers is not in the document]
leo-mazzone: https://uktrade.atlassian.net/browse/MTCHBX-462?atlOrigin=eyJpIjoiMjQ0Y2I5ZDZlMjgwNDUwMjgzNWU4NmFjMjMwMDk5ZDAiLCJwIjoiaiJ9

test/server/api/routes/test_routes_eval.py
leo-mazzone: literally doesn't work. Tried with `judgements.validate(full=False)`

Look at the first comment here https://stackoverflow.com/q/78848723

test/common/test_eval.py
leo-mazzone: however, clusters is more generic than expansion in general - expansion illustrates that expansion always comes with judgements and only contains the clusters that are relevant to it. In this instance, I moved expansion up and I renamed it to `judgement_cluster_expansion`. I've done these in both instances in this test

src/matchbox/common/eval.py
```diff
@@ -0,0 +1,227 @@
...
+        # (23), (24) and (34) will all be initially subtacted 1/4, and thus need to be
+        # then added 1 + 1/4 = 1 + (proportion of cluster endorsed in other rows)
+        validation_net_count.update(
+            {
+                p: validation_net_count.get(p, 0)
+                + 1
+                + ((len(shown) - len(endorsed)) / len(shown))
+                for p in positive_pairs
+            }
+        )
```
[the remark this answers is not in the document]
leo-mazzone: I agree this could be more legible, which is why I added the comment. Unfortunately, the set comprehension + single update operation is much more efficient than a for loop. Moving to different function harms locality of behaviour. Instead, I have adopted your positive and negative constant variables

src/matchbox/client/eval/mock_ui.py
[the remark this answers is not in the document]
will-langdale: That's not true, [Streamlit contains unit testing components](https://docs.streamlit.io/develop/api-reference/app-testing). I don't want to add Streamlit unit testing here though, as it only exists as an MVP thing and we may well flesh it out into something else (Plotly Dash for async, a CLI, etc etc).

src/matchbox/client/eval/mock_ui.py
```diff
@@ -0,0 +1,114 @@
...
+from matchbox.client.models.linkers import DeterministicLinker
+from matchbox.common.factories.sources import source_from_tuple
+from matchbox.common.graph import DEFAULT_RESOLUTION
+
+MOCK_WH_FILE = "sqlite:///eval_mock.db"
+
+
+def setup_mock_database():
+    """Add some mock data to test the evaluation UI."""
+    print("Mocking DB")
```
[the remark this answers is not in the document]
will-langdale: This isn't done -- have you pushed it?

src/matchbox/client/eval/ui.py
will-langdale: Not visible to me here. Need to push?

src/matchbox/server/postgresql/adapter.py
```diff
@@ -513,3 +552,296 @@ def delete_resolution(self, name: ResolutionName, certain: bool = False) -> None
...
+            ):
+                return user_id
+
+            user = Users(name=user_name)
+            session.add(user)
+            session.commit()
+
+            return user.user_id
+
+    def insert_judgement(self, judgement: CommonJudgement) -> None:  # noqa: D102
```
will-langdale, line 569: Can we factor out the logic, it's very long.

src/matchbox/server/postgresql/adapter.py
```diff
@@ -513,3 +552,296 @@ def delete_resolution(self, name: ResolutionName, certain: bool = False) -> None
...
+                        user_id=judgement.user_id,
+                        shown_cluster_id=judgement.shown,
+                        endorsed_cluster_id=endorsed_cluster_id,
+                        timestamp=datetime.now(timezone.utc),
+                    )
+                )
+
+                session.commit()
+
+    def get_judgements(self) -> tuple[Table, Table]:  # noqa: D102
```
will-langdale, line 629: Can we factor out the logic, it's very long.

src/matchbox/server/postgresql/orm.py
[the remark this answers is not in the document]
will-langdale: > Deleting the model doesn't currently delete the cluster

🤦 of course it doesn't

Yes, this is fine.

src/matchbox/server/postgresql/orm.py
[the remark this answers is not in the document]
will-langdale: > We don't want to keep a judgement without a user identity

Why?

It feels like we're working from different assumptions about why we'd need to delete users and judgements. I think we should leave this how it is and let the next step on how we manage this emerge from user behaviour and needs.

Feel free to resolve once read.

environments/development.env
```diff
@@ -29,3 +29,5 @@ MB__CLIENT__TIMEOUT=10
 MB__CLIENT__API_KEY=matchbox-api-key
 MB__CLIENT__RETRY_DELAY=2
 MB__CLIENT__DEFAULT_WAREHOUSE=
+MB__CLIENT__USER=scott.mcgregor
+MB__CLIENT__EVAL_RESOLUTION=
```
will-langdale, line 33: Perhaps?

```suggestion
MB__CLIENT__DEFAULT_EVAL_RESOLUTION=
```

Or the same as the default resolution, which should also be populated from here. But that's a bigger PR.

Or a `st.text_input()` and it's up to you to enter the right thing. This is my favoured edit to this PR.

Really we want a `st.dropdown()` populated from a list resolutions endpoint. This is my favoured actual solution in a new ticket.

src/matchbox/common/eval.py
```diff
@@ -0,0 +1,227 @@
...
+            (
+                len(true_positive_pairs) / len(model_pairs),
+                len(true_positive_pairs) / len(validation_pairs),
+            )
+        )
+
+    return pr_scores
+
+
+def process_judgements(
```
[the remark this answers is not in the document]
will-langdale: Clarity is so important here, I've had a really hard time. Let's get it right in this PR and pick optimisation up when it becomes a problem.

src/matchbox/common/exceptions.py
```diff
@@ -114,10 +135,24 @@ def __init__(self, message: str | None = None):
...
+
+        super().__init__(message)
+        self.user_id = user_id
+
+
 class MatchboxResolutionNotFoundError(Exception):
     """Resolution not found."""
 
-    def __init__(self, message: str | None = None, name: ResolutionName | None = None):
+    def __init__(self, message: str | None = None, name: str | None = None):
```
will-langdale, line 155: Why did you untype this?

will-langdale (commented):
My apologies for the belated comments on `adapter.py`'s factoring out, I overlooked how long that was getting last time. I also missed what was being stripped out of exceptions.

I was going over your PR comment and I realise I'd like to see a "create validation guide" ticket in the backlog please.

Other than that, this is very much getting there! Most comments are more or less continuations of existing discussions.

src/matchbox/common/exceptions.py
leo-mazzone: Because of a circular import. Something like, importing ResolutionName means importing DTOs, DTOs imports common.arrow, and common.arrow imports exceptions

environments/development.env
leo-mazzone: Add a way of setting this without an env variable

src/matchbox/server/postgresql/adapter.py
```diff
@@ -513,3 +552,296 @@ def delete_resolution(self, name: ResolutionName, certain: bool = False) -> None
...
+
+            # Cast to align with data transfer schema
+            final_samples = final_samples.set_column(
+                0, "root", final_samples["root"].cast(pa.uint64())
+            )
+            final_samples = final_samples.set_column(
+                1, "leaf", final_samples["leaf"].cast(pa.uint64())
+            )
+
+        return final_samples
```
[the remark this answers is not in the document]
leo-mazzone: Done

src/matchbox/common/eval.py
```diff
@@ -0,0 +1,230 @@
...
+
+    if len(expanded_judgements) != len(judgements):
+        raise ValueError("Malformed judgements / expansion data received.")
+
+    validation_net_count: dict[Pair, float] = {}
+    for shown, endorsed in zip(
+        expanded_judgements["shown_leaves"].to_list(),
+        expanded_judgements["endorsed_leaves"].to_list(),
+        strict=True,
+    ):
```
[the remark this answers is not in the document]
leo-mazzone: Done

environments/development.env
leo-mazzone: Done

PR #158, opened 2025-05-12
Resolution names are made the key entry point of resolution selection

ghost:
There are several concepts in Matchbox that are about to be reengineered in a more agnostic way. A single relational table is no longer the centre of the world.

* `resolution_name`, `model_name`, etc standardised to either:
    * `name`, if it's the only argument in a function that accesses only that resolution type
    * Its role in the function, such as `target`, `resolution`, `source` etc
* These arguments all type hinted with string type aliases

### ⚠️ Logic changes

Unlike the previous renaming PR, this one changes logic.

* `Resolutions.name` merges `SourceConfig.resolution_name` and `Resolution.resolution_name` and is now the key unique field
    * Trying to insert the same resolution name -- even with an amended description -- now errors. This paves the way for @leo-mazzone 's update routes
* `Resolutions.content_hash` and `Resolutions.resolution_hash` merge into `Resolutions_hash`, which is now purely the content hash and now DOESN'T need to be unique, as the name is

## 👀 Guidance to review

Make sure you're happy with the logic changes.

Confirm you're happy with changing `insert_model` taking `model_metadata` as its arg (rather than model), intended as a mirror to `source_config` as the argument for inserting a source.

Note I'm aware there's a strong chance the Alembic downgrade script would get the hash wrong because the original hash is ordered by data we don't have access to in the migration script. We have so few dependent users right now I don't think it matters.

## 🤖 AI declaration

Heavy usage in the Alembic, especially the downgrade recursive hash function, because it's annoyingly complex with relatively little importance or impact right now. Unit tests passing.

## 🔗 Relevant links

None.

## ✅ Checklist:

### DIFF
--- a/test/client/test_extract.py
+++ b/test/client/test_extract.py
@@ -94,7 +94,7 @@ def test_primary_keys_map(
         primary_keys_map(resolution="companies", engine=create_engine("postgresql://"))
 
     with pytest.raises(MatchboxSourceNotFoundError):
-        primary_keys_map(resolution_name="companies", full_names=["nonexistent"])
+        primary_keys_map(resolution="companies", full_names=["nonexistent"])
 
     # Case 1: apply engine filter, and retrieve single table
     foo_mapping = primary_keys_map(resolution="companies", engine=sqlite_warehouse)


src/matchbox/common/factories/dags.py
```diff
@@ -95,28 +106,30 @@ def add_model(self, testkit: ModelTestkit):
...
         Returns:
             A dictionary mapping:
             - LinkedSourcesTestkit keys (or None) to sets of model's source names
         """
-        if model_name not in self.models:
+        if name not in self.models:
             return {None: set()}
 
-        root_source_addresses = self.root_source_addresses.get(model_name, set())
+        root_source_addresses = self.root_source_addresses.get(name, set())
```
leo-mazzone, line 132: I've always found all these names confusing from the point of view of testkits. We're now saying that all names always map to a resolution name, right?

src/matchbox/common/factories/entities.py
```diff
@@ -192,10 +193,12 @@ def protected_names(cls, value: str) -> str:
 class EntityReference(frozendict):
     """Reference to an entity's presence in specific sources.
 
-    Maps dataset names to sets of primary keys.
+    Maps source names to sets of primary keys.
```
leo-mazzone, line 196: similar comment as above. Source names?

src/matchbox/server/base.py
```diff
@@ -250,7 +255,7 @@ def match(
         source_pk: str,
         source: SourceAddress,
         targets: list[SourceAddress],
-        resolution_name: str,
+        resolution: ResolutionName,
```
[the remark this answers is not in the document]
leo-mazzone: or even in this file actually

test/server/api/routes/test_routes_main.py
```diff
@@ -524,8 +524,8 @@ def test_api_key_authorisation(test_client: TestClient):
         (test_client.post, "/upload/upload_id"),
         (test_client.post, "/sources"),
         (test_client.post, "/models"),
-        (test_client.patch, "/models/model_name/truth"),
-        (test_client.delete, "/resolutions/model_name"),
+        (test_client.patch, "/models/name/truth"),
+        (test_client.delete, "/resolutions/name"),
```
leo-mazzone, line 528: this hurts me. Looking forward to PATCH /resolutions/name/truth
Feel free to resolve this comment

src/matchbox/server/postgresql/utils/insert.py
```diff
@@ -113,30 +114,27 @@ def insert_dataset(
 ) -> None:
```
leo-mazzone, line 114: insert_source?

src/matchbox/server/postgresql/utils/insert.py
```diff
@@ -298,62 +295,43 @@ def insert_dataset(
...
     """
-    log_prefix = f"Model {model}"
+    log_prefix = f"Model {name}"
     logger.info("Registering", prefix=log_prefix)
     with MBDB.get_session() as session:
-        resolution_hash = hash_values(
-            left.resolution_hash,
-            right.resolution_hash,
-            bytes(model, encoding="utf-8"),
+        model_name_hash = hash_values(
```
leo-mazzone, line 318: Don't be a fool Will

src/matchbox/server/postgresql/utils/query.py
```diff
@@ -106,7 +106,7 @@ def _union_valid_clusters(lineage_thresholds: dict[int, float]) -> Select:
 
     for resolution_id, threshold in lineage_thresholds.items():
         if threshold is None:
-            # This is a dataset - get all its clusters through ClusterSourcePK
+            # This is a source - get all its clusters through ClusterSourcePK
```
leo-mazzone, line 109: a source resolution

PR #161, opened 2025-05-14
Renamed SourceColumns to SourceFields throughout the codebase.

ghost:
There are several concepts in Matchbox that are about to be reengineered in a more agnostic way. A single relational table is no longer the centre of the world.

This means that "columns", a SQL concept, are better understood now as "fields". This performs this rename, but leaves mentions of columns in appropriate circumstances, such as in SQL tables and in dataframes.

For `SourceConfig`, this implements the `.index_fields` attribute.

* Rename columns to fields where appropriate
* Both key and indexed fields in `SourceConfig` are now `SourceField` objects, priming us for mapping
* Refactored `SourceConfigs` and `SourceFields` in the ORM
    * To/from DTO moved to ORM
    * Moved whether it's a key field to the fields

## 👀 Guidance to review

Nothing special.

## 🤖 AI declaration

Heavy use on Alembic migrations.

## 🔗 Relevant links

None.

## ✅ Checklist:

### DIFF
--- a/src/matchbox/common/db.py
+++ b/src/matchbox/common/db.py
@@ -71,7 +71,7 @@ def sql_to_df(
             Default is False.
         batch_size (int | None): Indicate the size of each batch when processing
             data in batches. Default is None.
-        schema_overrides (dict[str, Any] | None): A dictionary mapping field names
+        schema_overrides (dict[str, Any] | None): A dictionary mapping column names
             to dtypes. Default is None.
         execute_options (dict[str, Any] | None): These options will be passed through
             into the underlying query execution method as kwargs. Default is None.
--- a/src/matchbox/common/hash.py
+++ b/src/matchbox/common/hash.py
@@ -79,57 +79,64 @@ def hash_values(*values: tuple[T, ...]) -> bytes:
     return hashed_vals.digest()
 
 
-def process_field_for_hashing(field_name: str, schema_type: pl.DataType) -> plx.Expr:
-    """Process a field for hashing based on its type.
+def process_column_for_hashing(column_name: str, schema_type: pl.DataType) -> plx.Expr:
+    """Process a column for hashing based on its type.
 
     Args:
-        field_name: The field name
-        schema_type: The polars schema type of the field
+        column_name: The column name
+        schema_type: The polars schema type of the column
 
     Returns:
-        A polars expression for processing the field
+        A polars expression for processing the column
     """
     if isinstance(schema_type, pl.Binary):
-        return pl.col(field_name).fill_null("\x00").bin.encode("hex").alias(field_name)
+        return (
+            pl.col(column_name).fill_null("\x00").bin.encode("hex").alias(column_name)
+        )
     elif isinstance(schema_type, pl.Struct):
         return (
-            pl.col(field_name).struct.json_encode().fill_null("\x00").alias(field_name)
+            pl.col(column_name)
+            .struct.json_encode()
+            .fill_null("\x00")
+            .alias(column_name)
         )
     elif isinstance(schema_type, pl.Object):
         return (
-            pl.col(field_name)
+            pl.col(column_name)
             .map_elements(lambda x: str(x), return_dtype=pl.Utf8)
             .fill_null("\x00")
-            .alias(field_name)
+            .alias(column_name)
         )
     else:
-        return pl.col(field_name).cast(pl.Utf8).fill_null("\x00").alias(field_name)
+        return pl.col(column_name).cast(pl.Utf8).fill_null("\x00").alias(column_name)
 
 
 def hash_rows(
-    df: pl.DataFrame, fields: list[str], method: HashMethod = HashMethod.XXH3_128
+    df: pl.DataFrame, columns: list[str], method: HashMethod = HashMethod.XXH3_128
 ) -> pl.Series:
     """Hash all rows in a dataframe.
 
     Args:
         df: The DataFrame to hash rows from
-        fields: The field names to include in the hash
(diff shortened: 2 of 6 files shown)


src/matchbox/client/models/dedupers/base.py
```diff
@@ -10,7 +10,7 @@
 class DeduperSettings(BaseModel):
     """A data class to enforce basic settings dictionary shapes."""
 
-    id: str = Field(description="A unique ID column in the table to dedupe")
+    id: str = Field(description="A unique ID field in the data to dedupe")
```
leo-mazzone, line 13: Almost certainly out of scope, but why do we need to specify this separately in the deduper? I guess it could reuse the key field?

src/matchbox/common/db.py
```diff
@@ -71,7 +71,7 @@ def sql_to_df(
             Default is False.
         batch_size (int | None): Indicate the size of each batch when processing
             data in batches. Default is None.
-        schema_overrides (dict[str, Any] | None): A dictionary mapping column names
+        schema_overrides (dict[str, Any] | None): A dictionary mapping field names
```
leo-mazzone, line 74: But this is a SQL to DF function :)

src/matchbox/common/db.py
```diff
@@ -143,7 +143,7 @@ def get_schema_table_names(full_name: str) -> tuple[str, str]:
 
 
 def fullname_to_prefix(fullname: str) -> str:
-    """Converts a full name to a prefix for column names."""
+    """Converts a full name to a prefix for field names."""
```
leo-mazzone, line 146: Will this still be used in your location PR? Just trying to understand the direction we're heading

src/matchbox/common/factories/entities.py
```diff
@@ -183,12 +150,21 @@ def add_variations(self, *rule: VariationRule) -> "FeatureConfig":
         )
 
     @field_validator("name", mode="after")
-    def protected_names(cls, value: str) -> str:
+    @classmethod
```
leo-mazzone, line 153: Aren't field validators also classmethods by default?

src/matchbox/common/hash.py
```diff
@@ -157,34 +159,34 @@ def hash_arrow_table(
     if df.height == 0:
         return b"empty_table_hash"
 
-    columns: list[str] = sorted(df.columns)
-    df = df.select(columns)
+    fields: list[str] = sorted(df.columns)
```
leo-mazzone, line 162: There's a bit of a decision to make about whether we talk about columns here because we're in the context of a Polars dataframe. It's not the end of the world to leave as is tho

src/matchbox/server/postgresql/orm.py
```diff
@@ -359,34 +395,76 @@ class SourceConfigs(CountMixin, MBDB.MatchboxBase):
...
             address=SourceAddress(
                 full_name=self.full_name, warehouse_hash=self.warehouse_hash
             ),
-            key_field=self.key_field,
-            columns=[
-                CommonSourceColumn(
-                    name=column.column_name,
-                    type=column.column_type,
+            key_field=CommonSourceField(
+                name=self.key_field.name, type=self.key_field.type
```
leo-mazzone, line 460: this is one place where we should enforce string, is there any other one?

leo-mazzone: waaaaait. I guess there is no harm in capturing the key field type. It's the same as the other fields: we will always cast to string for you, we just need to know what the data originally was so that others can get it in the right format. So I guess, ignore the comment above?

src/matchbox/server/postgresql/orm.py
```diff
@@ -330,14 +341,39 @@ class SourceConfigs(CountMixin, MBDB.MatchboxBase):
...
+    )
+    key_field = relationship(
+        "SourceFields",
+        primaryjoin=(
+            "and_(SourceConfigs.source_config_id == SourceFields.source_config_id, "
+            "SourceFields.is_key == True)"
+        ),
+        viewonly=True,
+        uselist=False,
+        post_update=True,  # Handle circular dependency
```
leo-mazzone, line 366: I thought there wasn't one anymore?

leo-mazzone: circular dependency that is

src/matchbox/server/postgresql/orm.py
```diff
@@ -259,31 +261,40 @@ def reserve_block(
...
+    field_id = Column(BIGINT, primary_key=True)
     source_config_id = Column(
         BIGINT,
         ForeignKey("source_configs.source_config_id", ondelete="CASCADE"),
         nullable=False,
     )
-    column_index = Column(INTEGER, nullable=False)
-    column_name = Column(TEXT, nullable=False)
-    column_type = Column(TEXT, nullable=False)
+    index = Column(INTEGER, nullable=False)
```
leo-mazzone, line 276: why do we need the index? I thought when we hashed we ordered by field name?

src/matchbox/common/sources.py
```diff
@@ -357,9 +362,9 @@ class SourceConfig(BaseModel):
     name: SourceResolutionName = Field(
         default_factory=lambda data: str(data["address"])
     )
-    key_field: str
-    # Columns need to be set at creation, or initialised with `.default_columns()`
-    columns: tuple[SourceColumn, ...] | None = None
+    key_field: SourceField
+    # Fields need to be set at creation, or initialised with `.default_fields()`
```
leo-mazzone, line 366: Index fields can

PR #152, opened 2025-05-02
Swap sources to use locations

ghost:
We've implemented the location system so we can abstract away having to interface with different types of locations, like file systems or novel database types (Graph, NoSQL).

This brings this system into Sources proper.

## 💡 To do

- [x] Initial addition of locations to sources
- [x] Testkits patched and their unit tests pass
- [x] Source unit tests pass
- [x] Testkits handle other locations more realistically (currently heavily assume one SQL table)
- [ ] ORM updated
- [ ] Migrations
- [ ] Adapter tests updated and pass
- [ ] Remove `SourceAddress` throughout codebase
- [ ] Documentation updated, especially the DAG guide

## 🛠️ Changes proposed in this pull request

* Refactors sources to use locations
* Adds locations to the testkit system and updates factories to use locations

## 👀 Guidance to review

This is going to be a big PR. I've started a draft to encourage incremental reviews.

We've made a decision to remove Python query functions that support 🐰 tourist users for now. We're just supporting 🐝 builders in the short term while we move to a more semantic world. See the [Collections writeup](https://uktrade.atlassian.net/wiki/spaces/CDL/pages/4788781057/Collections) for user definitions.

## 🤖 AI declaration

In the current work, heavy, unmoderated use in the (failing) source unit tests.

## 🔗 Relevant links

* Ideation of locations in the [Collections writeup](https://uktrade.atlassian.net/wiki/spaces/CDL/pages/4788781057/Collections)

## ✅ Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] New and existing unit tests pass locally with my changes
- [ ] I've changed or updated relevant documentation

PR #166, opened 2025-05-16
Swap sources to use locations, redux

ghost:
We've implemented the location system so we can abstract away having to interface with different types of locations, like file systems or novel database types (Graph, NoSQL).

We've also done extensive renaming to shrink the size of the PR as small as possible.

This brings this system into Sources proper.

Closes #152 

- [x] Swap `SourceConfig` to use locations
- [x] Add locations to testkit subsystem
- [x] Update the ORM to use locations
- [x] Update helper functions
- [x] Remove `SourceAddress` from codebase
- [x] Update documentation

## 👀 Guidance to review

Are you okay with how I've changed selectors? We made a decision that we're explicitly making Python functions serve pipeline builders only for now.

Note my use of `URL._replace` in testkits is [lifted from the SQLAlchemy documentation](https://docs.sqlalchemy.org/en/20/core/engines.html#sqlalchemy.engine.URL.set), even though it's a private method.

## 🤖 AI declaration

Helped with Alembic migrations.

## 🔗 Relevant links

None.

## ✅ Checklist:

### DIFF
--- a/src/matchbox/common/factories/sources.py
+++ b/src/matchbox/common/factories/sources.py
@@ -7,15 +7,15 @@ from typing import Any, Callable, ParamSpec, TypeVar
 from unittest.mock import Mock, create_autospec
 
 import pandas as pd
+import polars as pl
 import pyarrow as pa
 from faker import Faker
 from pydantic import BaseModel, ConfigDict, Field
 from sqlalchemy import Engine, create_engine
-from sqlglot import select
+from sqlglot import cast, select
 from sqlglot.expressions import column
 
 from matchbox.common.arrow import SCHEMA_INDEX
-from matchbox.common.db import strip_credentials_from_uri, strip_driver_from_uri
 from matchbox.common.dtos import DataTypes, SourceResolutionName
 from matchbox.common.factories.entities import (
     ClusterEntity,
@@ -29,7 +29,6 @@ from matchbox.common.factories.entities import (
 )
 from matchbox.common.hash import hash_values
 from matchbox.common.sources import (
-    Location,
     RelationalDBLocation,
     SourceConfig,
     SourceField,
@@ -69,57 +68,6 @@ def make_features_hashable(func: Callable[P, R]) -> Callable[P, R]:
     return wrapper
 
 
-def create_single_table_location_resources(
-    key_field: SourceField,
-    index_fields: list[SourceField],
-    table_name: str,
-) -> tuple[str, Callable[[pa.Table, RelationalDBLocation, Engine], None]]:
-    """Create SQL and writer function for a single table.
-
-    A helper function to generate SQL and a writer for a RelationalDBLocation.
-    All fields are written to a single table with the specified name.
-
-    Args:
-        key_field: The identifier field (typically 'pk').
-        index_fields: List of SourceField objects to use as index fields.
-        table_name: Name of the table to create.
-
-    Returns:
-        A tuple containing:
-        - SQL query string to retrieve data from the single table
-        - Writer function that takes a PyArrow table, location, and credentials
-    """
-    all_field_names = [key_field.name] + [field.name for field in index_fields]
-
-    sql = select(*[column(name) for name in all_field_names]).from_(table_name).sql()
-
-    def _write_single_table(
-        data: pa.Table, location: RelationalDBLocation, credentials: Engine
-    ) -> None:
-        """Write all data to a single table."""
-        stripped_creds = strip_credentials_from_uri(
-            strip_driver_from_uri(str(credentials.url))
-        )
-        stripped_location = strip_driver_from_uri(str(location.uri))
-        if stripped_creds != stripped_location:
-            raise ValueError(
-                "The credentials provided do not match the location URI."
-                f"\nCredentials: {stripped_creds}"
-                f"\nLocation: {stripped_location}"
-            )
-
-        df = data.to_pandas()
-
-        df[all_field_names].to_sql(
-            name=table_name,
-            con=credentials,
-            index=False,
-            if_exists="replace",
-        )
-
-    return sql, _write_single_table
-
-
 class SourceTestkitParameters(BaseModel):
     """Configuration for generating a source."""
 
@@ -151,12 +99,6 @@ class SourceTestkit(BaseModel):
     data_hashes: pa.Table = Field(description="A PyArrow table of hashes for the data.")
     entities: tuple[ClusterEntity, ...] = Field(
         description="ClusterEntities that were generated from the source."
-    )
-    location_writer: Callable[[pa.Table, Location], None] = Field(
-        description=(
-            "A function that takes the data and location and writes the data to "
-            "the location."
-        )
     )
 
     @property
@@ -208,7 +150,11 @@ class SourceTestkit(BaseModel):
             set_credentials: Whether to set the credentials on the SourceConfig.
                 Offered here for convenience as it's often the next step.
         """
-        self.location_writer(self.data, self.source_config.location, credentials)
+        pl.from_arrow(self.data).write_database(
+            table_name=self.source_config.name,
+            connection=credentials,
+            if_table_exists="replace",
+        )
         if set_credentials:
             self.set_credentials(credentials)
 
@@ -309,8 +255,8 @@ class LinkedSourcesTestkit(BaseModel):
             set_credentials: Whether to set the credentials on the SourceConfig.
                 Offered here for convenience as it's often the next step.
         """
-        for source in self.sources.values():
-            source.write_to_location(credentials, set_credentials)
+        for source_testkit in self.sources.values():
+            source_testkit.write_to_location(credentials, set_credentials)
 
 
 def generate_rows(
@@ -612,21 +558,16 @@ def source_factory(
         SourceField(name=feature.name, type=feature.datatype) for feature in features
     )
 
-    # Create location, extract transform logic, and location writer function
-    location = RelationalDBLocation(
-        uri=str(engine.url._replace(username=None, password=None))
-    )
-    extract_transform, location_writer = create_single_table_location_resources(
-        key_field=key_field,
-        index_fields=index_fields,
-        table_name=name,
-    )
-
     # Create source config
     source_config = SourceConfig(
-        location=location,
+        location=RelationalDBLocation(uri=str(engine.url)),
         name=name,
-        extract_transform=extract_transform,
+        extract_transform=select(
+            cast(column(key_field.name), "string").as_(key_field.name),
+            *[column(field.name) for field in index_fields],
+        )
+        .from_(name)
+        .sql(),
         key_field=key_field,
         index_fields=index_fields,
     )
@@ -637,7 +578,6 @@ def source_factory(
         data=data,
         data_hashes=data_hashes,
         entities=tuple(sorted(results_entities)),
-        location_writer=location_writer,
     )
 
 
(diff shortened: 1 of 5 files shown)


src/matchbox/client/helpers/selector.py
```diff
@@ -1,78 +1,128 @@
...
+    def ensure_credentials(cls: type[Self], source: SourceConfig) -> SourceConfig:
+        """Ensure that the source has credentials set."""
+        if not source.location.credentials:
+            raise ValueError("Source credentials are not set")
+        return source
+
+    @property
+    def qualified_identifier(self) -> str:
+        """Get the qualified identifier for the source."""
+        return self.source.name + "_" + self.source.identifier
```
leo-mazzone, line 39: what is self.source.identifier? Do you mean self.source.key_field.name?

src/matchbox/client/helpers/selector.py
```diff
@@ -1,78 +1,128 @@
...
+    source: SourceConfig
+    fields: list[SourceField]
+
+    @field_validator("source", mode="after")
+    @classmethod
+    def ensure_credentials(cls: type[Self], source: SourceConfig) -> SourceConfig:
+        """Ensure that the source has credentials set."""
+        if not source.location.credentials:
+            raise ValueError("Source credentials are not set")
+        return source
```
leo-mazzone, line 34: why do all these methods exist at the selector level instead of the source config level?

src/matchbox/client/helpers/selector.py
```diff
@@ -1,78 +1,128 @@
...
+        """Get the qualified fields for the source."""
+        return [self.source.name + "_" + field.name for field in self.fields]
+
+    @classmethod
+    def from_name_and_credentials(
+        cls: type[Self],
+        name: SourceResolutionName,
+        credentials: Any,
+        fields: list[str] | None = None,
+    ) -> "Selector":
```
leo-mazzone, line 52: I question the usefulness of this method

src/matchbox/client/helpers/selector.py
```diff
@@ -1,78 +1,128 @@
...
 
 def select(
-    *selection: SourceResolutionName | dict[SourceResolutionName, str],
-    engine: Engine | None = None,
+    *selection: SourceResolutionName | dict[SourceResolutionName, list[str]],
+    credentials: Any | None = None,
 ) -> list[Selector]:
     """From one engine, builds and verifies a list of selectors.
 
+    Can we used on any number of sources as long as they share the same credentials.
```
leo-mazzone, line 78: can be used

src/matchbox/client/helpers/selector.py
```diff
@@ -1,78 +1,128 @@
...
+
     Args:
-        selection: Full source resolution names and optionally a subset of fields
-            to select
-        engine: The engine to connect to the data warehouse hosting the source.
-            If not provided, will use a connection string from the
-            `MB__CLIENT__DEFAULT_WAREHOUSE` environment variable.
+        selection: The source relations to retrieve data from
```
leo-mazzone, line 81: resolutions

leo-mazzone: or did you mean it?

src/matchbox/client/helpers/selector.py
```diff
@@ -84,77 +134,39 @@ def _process_query_result(
...
     Returns:
         The processed table with joined matchbox IDs and filtered fields
     """
     # Join data with matchbox IDs
     joined_table = data.join(
         other=mb_ids,
-        left_on=selector.address.format_field(key_field.name),
-        right_on="key",
+        left_on=selector.qualified_identifier,
+        right_on="source_identifier",
```
leo-mazzone, line 152: as above, thought we were calling these key fields now?

src/matchbox/client/helpers/selector.py
```diff
@@ -1,78 +1,128 @@
...
+            name: The name of the source to select from
+            credentials: The credentials to use for the source
+            fields: A list of fields to select from the source
+        """
+        source = _handler.get_source_config(name=name)
+        selected_fields: list[SourceField] = []
+        if fields:
+            for index_field in source.index_fields:
+                if index_field.name in fields:
+                    selected_fields.append(index_field)
```
leo-mazzone, line 65: the handling of un-indexed fields seem to have regressed. If you create a selector from __init__, you can give it whatever fields you want. If you use this factory function, the fields you pass it will be silently ignored

src/matchbox/server/postgresql/orm.py
```diff
@@ -391,7 +392,16 @@ def name(self) -> str:
...
+        UniqueConstraint(
+            "location_type",
+            "location_uri",
+            "extract_transform",
+            name="unique_location_transform",
+        ),
+        CheckConstraint(
+            "location_type IN ('rdbms')",
+            name="location_type_constraints",
+        ),
```
leo-mazzone, line 404: I think I disagree with this constraint at the ORM level

leo-mazzone: In fact I disagree with both

test/e2e/test_e2e.py
```diff
@@ -122,7 +122,9 @@ def setup_environment(
 
         # Setup code - Create tables in warehouse
         for source_testkit in self.linked_testkit.sources.values():
-            source_testkit.to_warehouse(engine=postgres_warehouse)
+            source_testkit.write_to_location(
+                credentials=postgres_warehouse, set_credentials=True
+            )
```
leo-mazzone, line 127: set_credentials=True is useful but will change the behaviour assumed in all the tests, you might have to look for all occurrences

docs/client/link-data.md
```diff
@@ -79,12 +96,15 @@ The `key_field` is the field in your source that contains some unique code that
 
 Each [`SourceConfig`][matchbox.common.sources.SourceConfig] object requires:
 
-- An `address` created with [`SourceAddress.compose()`][matchbox.common.sources.SourceAddress], comprised of
-    - The schema-qualified `full_name` of the dataset
-    - The `engine` used to connect
+- A `location`, which is a subclass of [`Location`][matchbox.common.sources.Location]. This will need a `type`, a `uri`, and `credentials`, the type of which changes depending on the type of location you're using
```
leo-mazzone, line 99: If all locations need this, why are we passing in the args to Locations.create as a dictionary?

leo-mazzone: Instead of args + kwargs?

test/client/test_extract.py
```diff
@@ -51,13 +53,13 @@ def test_key_field_map(
         return_value=Response(
             200,
             json=[
-                foo.source_config.model_dump(),
-                bar.source_config.model_dump(),
+                foo.source_config.model_dump(mode="json"),
```
leo-mazzone, line 56: why have you started dumping json explicitly? Just out of curiosity

test/client/test_helpers.py
```diff
@@ -454,59 +452,29 @@ def test_query_404_source_query(matchbox_api: MockRouter, sqlite_warehouse: Engi
         )
     )
 
-    sels = select({"foo": ["a", "b"]}, engine=sqlite_warehouse)
+    sels = select({"foo": ["crn", "company_name"]}, credentials=sqlite_warehouse)
 
     # Test with no optional params
     with pytest.raises(MatchboxSourceNotFoundError, match="42"):
```
leo-mazzone, line 458: what's a MatchboxSourceNotFoundError at the query level? Isn't it the same as the one above for resolution?

test/client/test_helpers.py
```diff
@@ -859,24 +747,35 @@ def test_match_404_resolution(matchbox_api: MockRouter, sqlite_warehouse: Engine
             ).model_dump(),
         )
     )
+    matchbox_api.get(f"/sources/{source_testkit.source_config.name}").mock(
```
leo-mazzone, line 750: why all this mocking source endpoints and writing to warehouse when previously it wasn't needed?

src/matchbox/common/sources.py
```diff
@@ -216,9 +186,7 @@ def validate_uri(cls, value: AnyUrl) -> AnyUrl:
             raise ValueError("Credentials should not be in the URI.")
```
leo-mazzone, line 186: I have seen somewhere a function to strip out credentials?

src/matchbox/common/sources.py
```diff
@@ -311,488 +279,309 @@ class SourceField(BaseModel):
...
+        """A helper method to detect the fields from the extract/transform logic.
 
-class SourceConfig(BaseModel):
-    """Data that can, or has been indexed on the backend.
+        Return a tuple of the key_field, and the index fields, as SourceFields.
+        """
+        id = key_field or "id"
+        df: pl.DataFrame = next(
+            location.execute(extract_transform=extract_transform, batch_size=100)
+        )
```
leo-mazzone, line 379: something feels circular here. I don't think we can detect the key and index fields separately from the extract transform

src/matchbox/common/sources.py
```diff
@@ -311,488 +279,309 @@ class SourceField(BaseModel):
...
-        """Custom equality which ignores engine."""
-        return (self.address, self.name, self.key_field, self.index_fields) == (
-            other.address,
-            other.name,
-            other.key_field,
-            other.index_fields,
-        )
+    @model_validator(mode="after")
+    def validate_key_field(self) -> Self:
+        """Ensure that the key field is a string."""
```
leo-mazzone, line 422: Not only. No harm in making docstrings more generic for something this small

src/matchbox/common/sources.py
```diff
@@ -311,488 +279,309 @@ class SourceField(BaseModel):
...
-            address=deepcopy(self.address, memo),
-            name=deepcopy(self.name, memo),
-            key_field=deepcopy(self.key_field, memo),
-            index_fields=deepcopy(self.index_fields, memo),
+    @model_validator(mode="after")
+    def validate_location_et_fields(self) -> "SourceConfig":
+        """Ensure that the location, extract_transform, and fields are aligned."""
+        if self.location.credentials is None:
+            # We can't validate
+            return self
```
leo-mazzone, line 436: Do we validate anywhere if we don't initially specify credentials?

test/client/test_extract.py
```diff
@@ -98,7 +98,7 @@ def test_key_field_map(
     with pytest.raises(MatchboxSourceNotFoundError):
         key_field_map(resolution="companies", uri_filter=["nonexistent"])
```
leo-mazzone, line 100: I'd recover case 0 in its earlier implementation so you check that the source filter actually does something

src/matchbox/common/sources.py
leo-mazzone: then why did you write that function?

docs/client/link-data.md
leo-mazzone: Could you expand? I see that the Location.create function is gone completely. How would we create a location of the right type from the a source config? Or I guess we wouldn't?

src/matchbox/common/factories/sources.py
```diff
@@ -227,6 +297,21 @@ def diff_results(
...
+
+        Credentials aren't set in testkits, so they must be provided here.
+
+        Args:
+            credentials: Credentials to use for the location.
+            set_credentials: Whether to set the credentials on the SourceConfig.
+                Offered here for convenience as it's often the next step.
+        """
+        for source in self.sources.values():
+            source.write_to_location(credentials, set_credentials)
```
leo-mazzone, line 313: why does the SourceConfig itself have this method? We don't really want to write to the location outside of tests

src/matchbox/common/factories/sources.py
```diff
@@ -62,13 +69,64 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
...
+        - SQL query string to retrieve data from the single table
+        - Writer function that takes a PyArrow table, location, and credentials
+    """
+    all_field_names = [key_field.name] + [field.name for field in index_fields]
+
+    sql = select(*[column(name) for name in all_field_names]).from_(table_name).sql()
+
+    def _write_single_table(
+        data: pa.Table, location: RelationalDBLocation, credentials: Engine
+    ) -> None:
```
leo-mazzone, line 98: Perplexed by this:

* why do you need this function instead of calling write_to_location?
* Why do you not just create the stripping behaviour implemented by the validation of a Location?
* Why do you use pandas instead of writing through polars?

src/matchbox/client/helpers/selector.py
leo-mazzone: Isn't the code in this method redundant given the selector validators?

leo-mazzone (approved):
Very excited about merging this

PR #169, opened 2025-05-20
Move hierarchical storage to root-leaf closure system

aidanrussell:
Adjacency lists are an antipattern for hierarchical storage ([SQL Antipatterns p.34](https://pragprog.com/titles/bksqla/sql-antipatterns/)). While the TAP team have suggested nested sets at a long term solution, we think a quick and dirty closure table is a stepping stone to give us the results we need, quickly.

This PR implements it.

* Adds a `Cluster` object to help work with hierarchy logic more easily, closes #168 
* Moves the ORM Contains table to root-leaf, and adds a flag to probabilities to recover pairs, closes #162 
* Changes `query()` to deal with new pattern, removing recursion
* Updates `insert_results()` to work with the new pattern, radically simplifying server-side wrangling
* Adds a destructive migration to move to the new ORM
* Adds a large amount of SQL tests for the PostgreSQL backend to help create efficient queries
* Removes a large amount of now-defunct code from transform, hash and insert modules
* Small change to `SplinkLinker` so that if multiple blocking rules produce multiple probabilities, we keep the highest. This is what the call to `.unique()` was supposed to do

Also removes large amounts of code that's no longer needed.

Some benchmarks based on my local tests:

* New `query()` returns in 6s
* New `match()` returns in 5.7ms

## 👀 Guidance to review

* Read the ORM. If you don't agree with that, raise it ASAP
* Review the `Cluster` object so you can understand what's going on with ingestion
* Review ingestion logic
* Review query and match logic

I'm not married to leaving in the SQL tests. I put them in to help me, and they did, hugely. I've explained in the docstring just how ephemeral they are, and how people should feel free to delete them the second they're not useful. The real coverage is from the backend tests.

## 🤖 AI declaration

I'm not going to mince words here: LLMs are incredibly adept and spanning the complex multilingual, multiparadigm systems that span a relational database, SQL queries, a Python ORM, and SQL queries written in Python. I make no apologies for relying on them to wrangle this work to its current stage. ✅  indicates this is LLM-influenced work that I've now checked over and am happy to endorse.

* ✅ Heavily used in the writing of unit tests for the `Cluster` object
* ✅ Heavily used in tidying the syntax of Polars statements, and ideating more efficient iterations of them
* ✅ Heavily used in writing and re-writing SQL queries in SQLAlchemy, as well as ideating about their decomposition
* ✅ Heavily used in generating the unit tests for SQL

## 🔗 Relevant links

* [SQL Antipatterns](https://pragprog.com/titles/bksqla/sql-antipatterns/)

## ✅ Checklist:


src/matchbox/common/hash.py
```diff
@@ -277,3 +277,126 @@ def has_mapping(self, *values: int) -> bool:
...
+        clusters = list(clusters)
+        if len(clusters) == 1:
+            return clusters[0]
+
+        intmap = intmap if intmap is not None else clusters[0]._intmap
+        unique_dict: dict[int, Cluster] = {}
+
+        for cluster in clusters:
+            if cluster.leaves is None:
+                unique_dict[id(cluster)] = cluster
```
leo-mazzone, line 392: any reason you're using the id instead of the hash? I don't mind either way, just curious

src/matchbox/common/hash.py
```diff
@@ -277,3 +277,126 @@ def has_mapping(self, *values: int) -> bool:
         """
         value_set = frozenset(values)
         return value_set in self.mapping
+
+
+class Cluster:
```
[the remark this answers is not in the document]
leo-mazzone: In fact, should this even be in common?

test/common/test_hash.py
```diff
@@ -257,3 +268,635 @@ def test_struct_json_hashing(method: HashMethod):
...
+            flag=test_flag.PAIR,
+        )
+
+        # Verify its properties
+        assert node.id == 1
+        assert node.hash == b"hash1"
+        assert node.leaves is None
+        assert node.flag == test_flag.PAIR
+
+    def test_add_leaves_to_disjoint_set(self, leaf_nodes: list[Cluster]):
```
leo-mazzone, line 347: what are all these DJS tests? It's not a common/hash.py class

leo-mazzone: In many places you're clearly testing behaviour that's out of scope. Some of these may be about interactions with Clusters? But actually I'm not convinced in general that any of these DJS tests validate anything about Clusters

src/matchbox/server/postgresql/orm.py
```diff
@@ -472,18 +472,18 @@ class Contains(CountMixin, MBDB.MatchboxBase):
     __tablename__ = "contains"
 
     # Columns
-    parent = Column(
+    root = Column(
```
leo-mazzone, line 475: I know you stubbornly call this root-leaf informally, but please not in the ORM. This is not necessarily a root, it's a parent

test/common/test_hash.py
```diff
@@ -257,3 +268,635 @@ def test_struct_json_hashing(method: HashMethod):
...
+        )
+
+        # Verify result has both flags set
+        assert result.flag == combined_flags
+        assert test_flag.PAIR in result.flag
+        assert test_flag.COMPONENT in result.flag
+        assert result.probability == 80
+        assert len(result.leaves) == 2
+
+    def test_combine_preserves_flag_from_input(
```
leo-mazzone, line 606: Fair bit of redundancy here too, in this test and the one below you're basically just testing that Cluster.combine doesn't ignore the flag you pass to it, we do not need to test this

test/common/test_hash.py
```diff
@@ -257,3 +268,635 @@ def test_struct_json_hashing(method: HashMethod):
...
+
+        # Verify the hash is the custom one
+        assert cluster.hash == custom_hash
+
+        # Verify it's different from what would have been calculated
+        auto_cluster = Cluster.combine([leaf_nodes[0], leaf_nodes[1]])
+        assert cluster.hash != auto_cluster.hash
+
+    def test_hash_generation_for_complex_hierarchy(
+        self, intmap: IntMap, leaf_nodes: list[Cluster]
```
leo-mazzone, line 734: you're not using the intmap

test/common/test_hash.py
```diff
@@ -257,3 +268,635 @@ def test_struct_json_hashing(method: HashMethod):
...
+        )
+
+        # Replace the first leaf with the modified one
+        modified_group = [modified_leaf] + leaf_nodes[1:]
+        modified_combined = Cluster.combine(modified_group)
+
+        # This time the hash SHOULD be different because we changed a leaf node
+        assert modified_combined.hash != original_hash
+
+    def test_scale_performance(self, intmap: IntMap):
```
leo-mazzone, line 814: this is fine to test this PR, but it doesn't feel right to keep this test around

src/matchbox/server/postgresql/utils/results.py
```diff
@@ -100,98 +99,114 @@ def get_model_config(resolution: Resolutions) -> ModelConfig:
...
+    if not pairwise_results:
+        # No pairwise clusters means no results to recover
+        return pa.table(
+            {
+                "id": pa.array([], type=pa.uint64()),
+                "left_id": pa.array([], type=pa.uint64()),
+                "right_id": pa.array([], type=pa.uint64()),
+                "probability": pa.array([], type=pa.uint8()),
+            }
+        )
```
leo-mazzone, line 149: as this is the same as the empty table above, could you use an inner function to generate it?

src/matchbox/server/postgresql/utils/results.py
```diff
@@ -100,98 +99,114 @@ def get_model_config(resolution: Resolutions) -> ModelConfig:
...
+
+    # 3. Build mapping from leaf cluster to parent cluster
+    # Each parent cluster "owns" certain leaf clusters
+    leaf_to_parent = {}
+    for parent_cluster_id, cluster_info in parent_clusters_dict.items():
+        for leaf_info in cluster_info["leaves"]:
+            leaf_id = leaf_info["leaf_id"]
+            leaf_to_parent[leaf_id] = parent_cluster_id
```
leo-mazzone, line 157: what does it mean for a cluster to own a leaf cluster? Of course, in general multiple clusters can point to the same leaf cluster

src/matchbox/server/postgresql/utils/results.py
```diff
@@ -100,98 +99,114 @@ def get_model_config(resolution: Resolutions) -> ModelConfig:
...
+
+    if not results:
+        return pa.table(
+            {
+                "id": pa.array([], type=pa.uint64()),
+                "left_id": pa.array([], type=pa.uint64()),
+                "right_id": pa.array([], type=pa.uint64()),
+                "probability": pa.array([], type=pa.uint8()),
+            }
         )
```
leo-mazzone, line 200: same here

src/matchbox/server/postgresql/utils/results.py
```diff
@@ -100,98 +99,114 @@ def get_model_config(resolution: Resolutions) -> ModelConfig:
...
+                        "left_id": parent_list[0],
+                        "right_id": parent_list[1],
+                        "probability": probability,
+                    }
                 )
-            ).label("children"),
-            func.min(Probabilities.probability).label("probability"),
+            elif len(parent_clusters_for_pair) == 1:
+                # Self-pair within same parent cluster - skip or handle differently
+                continue
```
leo-mazzone, line 187: what's a self pair and when does it occur? What does "handle differently" mean?

src/matchbox/server/postgresql/utils/results.py
```diff
@@ -100,98 +99,114 @@ def get_model_config(resolution: Resolutions) -> ModelConfig:
...
+            leaves_query = select(Contains.leaf).where(Contains.root == pair_cluster_id)
+            leaves = [row[0] for row in session.execute(leaves_query).all()]
+
+            # Find which parent clusters these leaves belong to
+            parent_clusters_for_pair = set()
+            for leaf_id in leaves:
+                if leaf_id in leaf_to_parent:
+                    parent_clusters_for_pair.add(leaf_to_parent[leaf_id])
+
+            # We expect exactly 2 parent clusters for a valid pair
```
leo-mazzone, line 174: what? why?

src/matchbox/server/postgresql/utils/results.py
leo-mazzone: I guess, is what you mean, from the point of view of a single resolution, there is only one path from a root to a leaf?

leo-mazzone: However, clusters at multiple levels from this resolution will point to the same leaves, how are you ensuring that in `leaf_to_parent` the values of the dict are actually the roots for the model?

src/matchbox/server/postgresql/utils/results.py
```diff
@@ -100,98 +99,114 @@ def get_model_config(resolution: Resolutions) -> ModelConfig:
...
+            leaf_id = leaf_info["leaf_id"]
+            leaf_to_parent[leaf_id] = parent_cluster_id
+
+    # 4. For each pairwise cluster, find which parent clusters it combines
+    results = []
+
+    with MBDB.get_session() as session:
+        for pair_cluster_id, probability in pairwise_results:
+            # Get leaves of this pairwise cluster
+            leaves_query = select(Contains.leaf).where(Contains.root == pair_cluster_id)
```
leo-mazzone, line 165: there's something that feels odd here. What you really need when you're trying to reconstruct pairwise probs isn't the leaves, it's the direct descendants of a pairwise cluster. Is this roundabout reconstruction plus the flag system not a bit of a smell that we need to store the original results in a `Results` table?

src/matchbox/server/postgresql/utils/insert.py
```diff
@@ -363,6 +377,211 @@ def _create_closure_entries(parent_resolution: Resolutions) -> None:
...
+                leaf = Cluster(id=leaf_id, hash=leaf_data["leaf_hash"], intmap=intmap)
+                cluster_lookup[leaf_id] = leaf
+            else:
+                leaf = cluster_lookup[leaf_id]
+
+            # Add leaf to cluster's leaves list
+            leaves.append(leaf)
+
+        cluster.leaves = tuple(leaves)
+        cluster_lookup[cluster_id] = cluster
```
leo-mazzone, line 424: this is unnecessary

src/matchbox/server/postgresql/utils/insert.py
```diff
@@ -363,6 +377,211 @@ def _create_closure_entries(parent_resolution: Resolutions) -> None:
...
+    # Look up existing clusters in the database
+    with ingest_to_temporary_table(
+        table_name="hashes",
+        schema_name="mb",
+        column_types={
+            "cluster_hash": BYTEA,
+        },
+        data=all_clusters_df.select("cluster_hash").to_arrow(),
+    ) as temp_table:
+        existing_cluster_stmt = select(Clusters.cluster_id, Clusters.cluster_hash).join(
```
leo-mazzone, line 541: since we don't index on bytes, this will be an expensive join. Could we restrict to those clusters within either the left or the right resolution?

src/matchbox/server/postgresql/utils/insert.py
```diff
@@ -402,102 +621,82 @@ def _results_to_insert_tables(
...
-    hierarchy_new = pl.from_arrow(hierarchy_new).unique().to_arrow()
 
-    contains = pa.table(
-        {
-            "parent": hm.generate_ids(hierarchy_new["parent"]),
-            "child": hm.generate_ids(hierarchy_new["child"]),
-        }
+    # Use all clusters and unnest probabilities and role flag for Probabilities table
+    new_probabilities_df = (
+        all_clusters_df.select("cluster_id", "cluster")
```
leo-mazzone, line 665: why all cluster? Some clusters will be superseded by the new clusters. Let's discuss

src/matchbox/server/postgresql/utils/insert.py
```diff
@@ -363,6 +377,211 @@ def _create_closure_entries(parent_resolution: Resolutions) -> None:
...
+
+        for leaf_data in data["leaves"]:
+            leaf_id = leaf_data["leaf_id"]
+
+            # Create leaf if it doesn't exist in lookup
+            if leaf_id not in cluster_lookup:
+                leaf = Cluster(id=leaf_id, hash=leaf_data["leaf_hash"], intmap=intmap)
+                cluster_lookup[leaf_id] = leaf
+            else:
+                leaf = cluster_lookup[leaf_id]
```
leo-mazzone, line 418: why would the leaf id ever be in the cluster lookup? Isn't the nested data only inclusive of the top-level clusters for the two parent resolutions?

src/matchbox/server/postgresql/utils/insert.py
```diff
@@ -363,6 +377,211 @@ def _create_closure_entries(parent_resolution: Resolutions) -> None:
...
+        probabilities: Arrow table containing probability data
+
+    Returns:
+        Dictionary mapping cluster hashes to Cluster objects
+    """
+    logger.debug("Computing hierarchies")
+
+    djs = DisjointSet[Cluster]()
+    all_clusters: dict[bytes, Cluster] = {}
+    component_cache: set[frozenset[Cluster]] = set()
```
leo-mazzone, line 469: could we call this "seen components"?

src/matchbox/server/postgresql/utils/insert.py
```diff
@@ -363,6 +377,211 @@ def _create_closure_entries(parent_resolution: Resolutions) -> None:
...
+            frozenset(component) for component in djs.get_components()
+        }
+        for component in components.difference(component_cache):
+            cluster = Cluster.combine(
+                clusters=component,
+                probability=probability,
+                flag=ClusterClosureRole.COMPONENT,
+            )
+            if cluster.hash in all_clusters:
+                all_clusters[cluster.hash].add_flag(ClusterClosureRole.COMPONENT)
```
leo-mazzone, line 484: If you're checking whether the hash exists already in the dictionary you're building, I don't think you need the component_cache at all?

src/matchbox/common/hash.py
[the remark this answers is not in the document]
will-langdale: On common, yes, I think this is an important object that can help us reason about clusters across the codebase. It's an attempt as a fundamental object.

On `hash.py`, I think yes, for now? I made it to help wrangle and hash. Would you prefer `transform.py`? That's a strong contender.

src/matchbox/common/hash.py
will-langdale: I honestly don't remember and won't change if it makes no difference.

src/matchbox/common/hash.py
```diff
@@ -277,3 +277,126 @@ def has_mapping(self, *values: int) -> bool:
...
+            intmap: An IntMap instance for generating unique IDs. Defaults to the
+                first cluster's IntMap if not provided.
+            probability: the probability of the cluster from its resolution
+            flag: A flag indicating the role of the cluster
+
+        Returns:
+            A new Cluster containing all unique leaves from the input clusters
+        """
+        if not clusters:
+            return None
```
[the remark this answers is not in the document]
will-langdale: Removed.

test/common/test_hash.py
```diff
@@ -257,3 +268,635 @@ def test_struct_json_hashing(method: HashMethod):
...
+            intmap=intmap, probability=None, leaves=None, id=1, hash=b"hash1"
+        )
+
+        # Verify its properties
+        assert node.id == 1
+        assert node.hash == b"hash1"
+        assert node.leaves is None
+        assert node.flag is None
+
+    def test_create_leaf_node_with_flag(self, intmap: IntMap, test_flag: Flag):
```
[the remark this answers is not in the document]
will-langdale: Done.

test/common/test_hash.py
```diff
@@ -257,3 +268,635 @@ def test_struct_json_hashing(method: HashMethod):
...
+        # Create a completely different structure with the same leaves
+        # This time using combine with all leaves
+        combined = Cluster.combine([leaf_nodes[0], leaf_nodes[1], leaf_nodes[2]])
+
+        # Should still have the same hash
+        assert direct.hash == combined.hash
+
+    def test_hash_with_provided_hash_value(
+        self, intmap: IntMap, leaf_nodes: list[Cluster]
+    ):
```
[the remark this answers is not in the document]
will-langdale: Because a universal `Cluster` object should read hashes out from an existing system, and work with them.

test/common/test_hash.py
will-langdale: Removed flags and all flag tests.

src/matchbox/server/postgresql/utils/results.py
```diff
@@ -100,98 +99,114 @@ def get_model_config(resolution: Resolutions) -> ModelConfig:
...
+                    }
                 )
-            ).label("children"),
-            func.min(Probabilities.probability).label("probability"),
+            elif len(parent_clusters_for_pair) == 1:
+                # Self-pair within same parent cluster - skip or handle differently
+                continue
+            else:
+                # Invalid pair (0 or >2 parents) - skip
+                continue
```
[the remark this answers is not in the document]
will-langdale: Results storage completely refactored.

src/matchbox/server/postgresql/utils/results.py
will-langdale: Results storage completely refactored.

src/matchbox/server/postgresql/utils/results.py
will-langdale: Results storage completely refactored.

test/common/test_hash.py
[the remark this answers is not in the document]
will-langdale: In fact just removed them both.

src/matchbox/server/postgresql/orm.py
will-langdale: Discussed.

test/common/test_hash.py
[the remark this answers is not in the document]
will-langdale: The opposite should be true: hash should be overwritten.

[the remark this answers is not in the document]
will-langdale: BUT we now have a new problem: a big reason you would want to provide a hash is that you already performed this expensive calculation once and don't want to do it again. Validating this difference throws away this benefit and will make ingest much slower.

This leaves this test in an odd position. The hash _should_ be overwritten, but in practise, we just need to trust that the hash we read out of the database is the hash that pertains to this object. Neither choice of which hash is the defining one -- loaded or calculated -- is computationally worth validating and testing, but we do need to privilege the one read out of the system to get the speedup.

I therefore think that this test is pointless, because this is a necessity of the implementation, not a feature, and I'm deleting it.

src/matchbox/server/postgresql/utils/insert.py
will-langdale: Simplified object into one pass.

src/matchbox/server/postgresql/utils/insert.py
will-langdale: `_build_cluster_objects()` has been streamlined.

src/matchbox/server/postgresql/utils/insert.py
```diff
@@ -363,6 +377,211 @@ def _create_closure_entries(parent_resolution: Resolutions) -> None:
...
+                    pl.arange(0, new_clusters_df.shape[0], dtype=pl.Int64)
+                    + next_cluster_id
+                ).alias("cluster_id"),
+                pl.lit(True).alias("new"),
+            ]
+        )
+    else:
+        new_clusters_df = new_clusters_df.with_columns(
+            [pl.lit(None).cast(pl.Int64).alias("cluster_id"), pl.lit(True).alias("new")]
+        )
```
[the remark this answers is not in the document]
will-langdale: You do but the logic can be tidied. `PKSpace.reserve_block()` errors if the number is less than 0. I've refocused the check on that issue specifically.

src/matchbox/server/postgresql/utils/insert.py
```diff
@@ -363,6 +292,198 @@ def _create_closure_entries(parent_resolution: Resolutions) -> None:
...
+            all_clusters[cluster.hash] = cluster
+
+        return components
+
+    for left_cluster, right_cluster, probability in _results_to_cluster_pairs(
+        cluster_lookup, probabilities
+    ):
+        if probability < threshold:
+            # Process the components at the previous threshold
+            seen_components = _process_components(probability)
```
leo-mazzone, line 395: Is this right? Should you not process components at threshold instead?

src/matchbox/server/postgresql/utils/insert.py
leo-mazzone: what if you see the same cluster with a different probability? I guess it can't happen. In which case, won't it be simpler to use the cluster hash instead of the whole cluster object? I don't even see a custom __hash__ implemented on Clusters, so how does it work?

src/matchbox/server/postgresql/utils/insert.py
will-langdale: Yes, great catch. This condition isn't caught in testing because everything is deterministic. Perhaps this is a reason #196 might be valuable: it's the only probabilistic testing in the suite right now.

leo-mazzone: could you add a small test with the simplest scenario to at least have something?

src/matchbox/server/postgresql/utils/query.py
```diff
@@ -43,222 +55,388 @@ def get_source_config(name: SourceResolutionName, session: Session) -> SourceCon
...
+        select(
+            target_cluster_cte.c.cluster_id.label("cluster"),
+            matching_leaves_cte.c.source_config_id,
+            matching_leaves_cte.c.key,
+        )
+        .select_from(
+            target_cluster_cte.outerjoin(
+                matching_leaves_cte,
+                literal(True),  # Always join - we want all target_cluster rows
             )
```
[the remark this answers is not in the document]
leo-mazzone: How does differ from the cross join happening in the matching_leaves_cte?

src/matchbox/server/postgresql/utils/query.py
```diff
@@ -267,265 +445,128 @@ def query(
...
-def _build_hierarchy_down(
-    highest_parent: Select, unnested_clusters: CTE, contains_table: CTE | Contains
-) -> CTE:
-    """Build recursive CTE that finds all child clusters and their IDs.
+    * Parent discovery: Queries ResolutionFrom to find all direct parent
+        resolutions (level 1)
+    * Cluster building: For each parent, runs the full unified query to get all
+        cluster assignments with both root and leaf information
+    * Aggregation: Collects all leaf nodes belonging to each root cluster across all
+        parent resolutions
```
leo-mazzone, line 516: I find this confusing. In parent discovery, it's not just level 1 we care about is it?

src/matchbox/server/postgresql/utils/query.py
```diff
@@ -267,265 +445,128 @@ def query(
...
+            # Execute and collect results
+            for row in session.execute(parent_assignments):
+                root_id = row.root_id
+                if root_id not in cluster_leaves:
+                    cluster_leaves[root_id] = set()
+                    all_clusters[root_id] = {
+                        "root_hash": row.root_hash,
+                        "leaves": [],
+                        "probability": None,
+                    }
```
leo-mazzone, line 559: the indentation looks wrong

src/matchbox/server/postgresql/utils/query.py
```diff
@@ -267,265 +445,128 @@ def query(
...
-        )
-        .distinct()
-        .select_from(hierarchy_down)
-    )
+        # Convert tuple sets to dict lists
+        for cluster_id, leaf_tuples in cluster_leaves.items():
+            all_clusters[cluster_id]["leaves"] = [
+                {"leaf_id": leaf_id, "leaf_hash": leaf_hash}
+                for leaf_id, leaf_hash in leaf_tuples
+            ]
```
leo-mazzone, line 567: why two distinct loops?

test/server/postgresql/test_pg_sql.py
```diff
@@ -0,0 +1,1501 @@
...
+        with MBDB.get_session() as session:
+            source_a_resolution = session.get(Resolutions, 1)  # source_a resolution
+            source_a_config = session.get(SourceConfigs, 11)  # source_a config
+
+            query = _build_unified_query(
+                resolution=source_a_resolution,
+                sources=[source_a_config],  # Pass source config directly
+                threshold=None,
+                columns=columns,
+            )
```
leo-mazzone, line 630: I thought that _build_unified_query was called when the resolution differed from the source?

test/server/postgresql/test_pg_sql.py
```diff
@@ -0,0 +1,1501 @@
...
+            * At 90%, pairwise component 501=301+401=101+102+201+202
+            * At 80%, pairwise 502=301+205=101+102+205, which forms component
+                503=301+205+401=101+102+201+202+205
+        * Cluster 504, which mixes raw leaves:
+            * At 90%, pairwise component 504=103+203 (103 undeduped at DA=80)
+
+    This diagram shows the structure of the test dataset. Each node shows the pairwise
+    clusters that would be formed by a model's results, but can also be traced back to
+    the sources to show the root-leaf relationships that are created. For example,
+    cluster 502 is stored only so we can recover pairs, because 503 is the true cluster.
```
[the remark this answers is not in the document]
will-langdale: Performed a really comprehensive overhaul of the docstrings and data, finding some serious flaws that further flush out our probabilities bug.

src/matchbox/server/postgresql/utils/insert.py
will-langdale: This led to a series of bugfixes which updated our benchmark times. Slower, but they're still excellent.

* New `query()` returns in 6.2s
* New `match()` returns in 5.7ms

test/server/postgresql/test_pg_sql.py
```diff
@@ -0,0 +1,1501 @@
...
+        """Should raise exception for nonexistent source."""
+        with pytest.raises(MatchboxSourceNotFoundError):
+            query("nonexistent", resolution=None)
+
+    def test_query_nonexistent_resolution(
+        self, populated_postgres_db: MatchboxPostgres
+    ):
+        """Should raise exception for nonexistent resolution."""
+        with pytest.raises(MatchboxResolutionNotFoundError):
+            query("source_a", resolution="nonexistent")
```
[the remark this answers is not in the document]
will-langdale: Removed.

src/matchbox/server/postgresql/utils/query.py
```diff
@@ -267,265 +445,128 @@ def query(
...
-
-def _find_highest_parent(hierarchy_up: CTE) -> Select:
-    """Find the topmost parent cluster from the hierarchy."""
-    return (
-        select(hierarchy_up.c.parent)
-        .order_by(hierarchy_up.c.level.desc())
-        .limit(1)
-        .scalar_subquery()
-    )
+def get_clusters_with_leaves(
```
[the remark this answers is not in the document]
will-langdale: Changed to `get_parent_clusters_and_leaves()` to be very explicit.

I moved it to `insert.py`, but then `_build_unified_query()` needs to become the public function, which feels more wrong to me. I personally feel that leaving it how it is is the best reflection of what should be public and private between `insert.py and `query.py`.

src/matchbox/server/postgresql/utils/query.py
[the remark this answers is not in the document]
will-langdale: I've simplified it to remove this second cross join. I know it could be simplified further but I'm really, really hesitant to make larger edits that would need testing at scale again, which takes me a long time to set up right now, and this is already very fast.

Would you be accept this edit + a ticket to simplify further?

src/matchbox/server/postgresql/orm.py
```diff
@@ -582,6 +583,7 @@ class Results(CountMixin, MBDB.MatchboxBase):
 
     # Constraints
     __table_args__ = (
-        CheckConstraint("probability BETWEEN 0 AND 100", name="valid_probability"),
         Index("ix_results_resolution", "resolution_id"),
+        CheckConstraint("probability BETWEEN 0 AND 100", name="valid_probability"),
+        UniqueConstraint("resolution_id", "left_id", "right_id"),
```
leo-mazzone, line 588: are left id and right id sorted?

leo-mazzone: or do they need to be?

test/server/test_adapter.py
```diff
@@ -313,6 +335,62 @@ def test_model_results_basic(self):
...
+                resolution="probabilistic_test.crn",
+            )
+            res_clusters = query_to_cluster_entities(
+                query=res,
+                keys={dag.sources["crn"].name: "key"},
+            )
+
+            identical, report = diff_results(
+                expected=dag.models["probabilistic_test.crn"].entities,
+                # expected=dag.linked["linked_cdms_crn_duns"].true_entity_subset("crn"),
```
leo-mazzone, line 354: can this line go?

test/fixtures/db.py
```diff
@@ -168,6 +168,51 @@ def create_dedupe_scenario(
...
+    warehouse_engine: Engine,
+    n_entities: int = 10,
+    seed: int = 42,
+) -> TestkitDAG:
+    """Create a dedupe TestkitDAG scenario."""
+    # First create the index scenario
+    dag = create_index_scenario(backend, warehouse_engine, n_entities, seed)
+
+    # Get the linked sources
+    linked_key = next(iter(dag.linked.keys()))
```
leo-mazzone, line 183: can I confirm there's only one?

src/matchbox/common/transform.py
```diff
@@ -242,6 +242,16 @@ def __init__(
...
 
+    def __hash__(self) -> int:
+        """Return a hash of the cluster based on its hash bytes."""
+        return hash(self.hash)
+
+    def __eq__(self, other: Any) -> bool:
+        """Check equality of two Cluster objects based on their hash."""
+        if not isinstance(other, Cluster):
+            return False
+        return self.hash == other.hash
```
leo-mazzone, line 253: is there any code now or in the future that might assume the probability is equal as well?

src/matchbox/server/postgresql/utils/query.py
leo-mazzone: that's fine, but this code implies that the same root_id can appear either across parents, or across rows of `parent_assignments`, why is this the case? Intuitively, the same root id points to the same leaves, so you only need to process the same root_id once, and yet `cluster_leaves` is added to on line 566 outside of the if statement

src/matchbox/server/postgresql/orm.py
will-langdale: I don't follow at all. Why would we sort them? This is basically a place to dump `Results.probabilities` with basic referential integrity.

src/matchbox/common/transform.py
will-langdale: Currently, the assumption that a cluster's leaves (and therefore its hash) define it and its equality, is a key part of how it's used logically. In that set loop, we are specifically seeing the same clusters are lower probabilities, but they must still be the same as the one we saw at the higher.

I can't speak for future code, but this type of identity is core to my long-term thinking about cluster objects (see [EntityFrame](https://github.com/will-langdale/entityframe)).

src/matchbox/server/postgresql/utils/query.py
will-langdale: I went down the rabbit hole, and discovered a few bugs.

* `_build_unified_query()` wasn't filtering to sources in the lineage when `source=None`, causing sources from outside the lineage to appear in this query. Fixed.
* `Resolutions.get_lineage()` needs to return source IDs to do this
* A single hash can have multiple keys. However, because we don't actually use the keys, we can drop the keys and source config ID columns from the "full" representation, and use `distinct()` to deduplicate server-side
* "full" now seems like an inappropriate literal. Replaced with "root_leaf"

src/matchbox/server/postgresql/orm.py
```diff
@@ -160,17 +173,25 @@ def get_lineage(
                     .all()
                 )
 
+                valid_parents = list(set(source_resolution_ids + descendant_ids))
```
leo-mazzone, line 176: why the set? Can there be overlap between valid source resolutions and their descendants?

test/server/postgresql/test_pg_sql.py
```diff
@@ -634,30 +636,35 @@ def test_build_unified_source_only(
...
             assert all(result["root_id"] == result["leaf_id"])
-            assert all(result["source_config_id"] == 11)
+
+            # Should have clusters 101, 102, 103, 104, 105
+            cluster_ids = set(result["root_id"])
+            assert cluster_ids == {101, 102, 103, 104, 105}
+
         else:  # id_key
+            # Should return all 6 keys from source_a
+            assert len(result) == 6
```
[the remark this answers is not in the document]
will-langdale: Renamed, added to docstring.

src/matchbox/server/postgresql/orm.py
will-langdale: We don't. There shouldn't be duplication and there's no performance or logic impact if there is. Removed.

leo-mazzone (approved):
This is a really strong implementation of what was a vast and tricky change. Well done.

PR #177, opened 2025-05-23
Add integration test for DAGs and fix all bugs that emerge

ghost:
Adds an e2e integration test for the DAG API and fixes all bugs.

* Adds a big integration test for the DAG API
* Fixed `StepInput` to work properly with select
* Fixed `_handler.get_model()` to be able to check if there's an existing model

Also:

* Minor change to delete model call argument
* Minor documentation tweaks overlooked in previous PRs

## 👀 Guidance to review

My only controversial decision, I think, is how I handle checking whether a model exists in the database. We are currently using a fast fail pattern where the 404 is immediately turned into an error by the event system. What I really need is to be able to handle that 404 as not an error, but a legitimate statement that the model isn't there.

For now I solve with `try ... except` but I don't like this pattern in the long run. The level of rearchitecting that would be needed to do otherwise is out of scope for this PR, I think.

Another possible controversial decision is allowing `ModelConfig`'s left and right resolution to be interchangeable. Again, would need some very fundamental changes database side to do this properly on retrieval.

## 🤖 AI declaration

Heavy usage in synthesising the existing unit tests and actual DAG code into a new e2e test. All read through and checked manually -- it did an excellent job, I think.

## 🔗 Relevant links

None.

## ✅ Checklist:


leo-mazzone (approved):
LGTM

PR #180, opened 2025-05-28
Tidy up Datadog

ghost:
I was playing with Datadog and wanted configuring it to be simpler with less files.

Note these changes only affect local use of Datadog, which we do very very little of anyway. Even more reason to keep it simple, imo.

* Datadog env files defined in docker compose
* Removed public env files, renamed private env files to make it simpler to use
* Explicitly defined container port

## 👀 Guidance to review

None.

## 🤖 AI declaration

None.

## 🔗 Relevant links

* [Code search](https://github.com/search?q=%22datadog%2Fagent%3A7%22+path%3Adocker-compose.yaml&type=code) showing most people defining environment variables in their compose too, on the first page at least

## ✅ Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,9 +21,6 @@ notebooks/tmp*
 # Documentation
 site/
 
-# Datadog
-datadog.env
-
 ##########################
 # Boilerplate .gitignore #
 ##########################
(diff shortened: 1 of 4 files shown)


leo-mazzone (changes requested):
I think this is confusing:

* Now the environments folder contains two different types of environments
* A non-DD environment would be copied to the root of the repo, not to the environments folder

PR #186, opened 2025-06-03
SourceConfig.new() transpiles SQL

ghost:
A user reported that when configuring a source, `SRING_AGG` became `GROUP_CONCAT`. I think this was because we used SQLGlot in `.new()`, which is wrong for two reasons.

1. Working out the dialect based on the SQLAlchemy engine would be needed to make the existing code work, and the names they use aren't 1:1
2. Sources could have any location, not just SQL-based ones, so SQLGlot was wrong here anyway

* Removes the code, adds a unit test

## 👀 Guidance to review

I know that this will mean `.new()` takes longer to run, but them's the breaks for now without a new location method.

## 🤖 AI declaration

Used to find the bug and find a good SQLite/SQLGlot crossover to reproduce. Code is all me.

## 🔗 Relevant links

* [Quirks, Caveats, and Gotchas In SQLite](https://www.sqlite.org/quirks.html)

## ✅ Checklist:

### DIFF
--- a/src/matchbox/common/sources.py
+++ b/src/matchbox/common/sources.py
@@ -29,7 +29,6 @@ from pydantic import (
 )
 from sqlalchemy import Engine
 from sqlalchemy.exc import OperationalError
-from sqlglot import parse_one
 
 from matchbox.common.db import (
     QueryReturnType,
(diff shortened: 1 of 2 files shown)


sophie-daintta (approved):
Looks good to me!

PR #197, opened 2025-06-17
Add debugging options to DAG

leo-mazzone:
We were finding it difficult to debug a DAG in production. This is a small set of changes that addresses different problems:

* Iterate quickly on tiny subsets of specific examples we want to test end-to-end. Similar to Splink adversarial examples, but for a whole DAG
* Allow to inspect model results and query source data

* Tidy up the debug arguments passed to the method to run DAG
* Allow to store in memory outputs of all DAG nodes for debugging: `Results` for models, and hashes for indexes
* Allow to override the content of a source at runtime, as long as the key field and index fields conform to the original `SourceConfig`
* Added simpler notation to qualify fields

## 👀 Guidance to review
Unsure about storing the hashes in the debug outputs:

* we are producing hashes twice, once in the handler and once in the DAG logic. I didn't want to have to rewire a bunch of things, and though inefficient, this should be fine when debugging
* not super sure how useful the hashes will be, but they let you see how many entries are being treated as unique source clusters

I have also not documented these options in e.g. a tutorial, however this interface feels unstable for now, I suggest we return to this later.

## 🤖 AI declaration

None.

## 🔗 Relevant links

None.

## ✅ Checklist:

- [x] All new code is tested
    - [ ] Tutorials
    - [ ] Developer docs


src/matchbox/client/dags.py
```diff
@@ -35,7 +48,7 @@ def inputs(self) -> list["StepInput"]:
         ...
 
     @abstractmethod
-    def run(self) -> None:
+    def run(self) -> Any:
```
will-langdale, line 51:
```suggestion
    def run(self) -> Table | Results:
```

src/matchbox/client/dags.py
```diff
@@ -224,6 +241,7 @@ def __init__(self):
         self.nodes: dict[ResolutionName, Step] = {}
         self.graph: dict[ResolutionName, list[ResolutionName]] = {}
         self.sequence: list[ResolutionName] = []
+        self.debug_outputs: dict[ResolutionName, Any] = {}
```
will-langdale, line 244:
```suggestion
        self.debug_outputs: dict[ResolutionName, Table | Results] = {}
```

src/matchbox/client/dags.py
```diff
@@ -387,51 +405,80 @@ def format_children(node: ResolutionName, prefix=""):
...
-                raise ValueError(f"Step {finish} not in DAG") from e
+                raise ValueError(f"Step {debug_options.finish} not in DAG") from e
         else:
             end_index = len(self.sequence)
 
+        self.debug_outputs.clear()
+
+        # Create debug warehouse if needed
+        if len(debug_options.override_sources):
+            debug_sqlite_uri = "sqlite:///debug_location.db"
```
will-langdale, line 444: You might want to organise these, save and load them, show them to your grandchildren, etc. Add a `Path` to the debug options imo.

will-langdale (changes requested):
Some minor changes.

Some things I'd love you to bring into scope:

* Report row and entity count when querying
* Report row count for results (would love entities too but needs connected componants)
* Report time taken per step ("step completed in x")
* Correct the formatting on this message. Should have `[]`:

```
2025-06-06 10:02:34 [matchbox] [Model debug_naive_data_hub] Retrieving metadata
2025-06-06 10:02:34 [matchbox] Model debug_naive_data_hub already exists. Passing.
2025-06-06 10:02:34 [matchbox] [Model debug_naive_data_hub] Uploading results
```

* Give some better logging on what's happening between "finished" and "retrieving metadata":

```
2025-06-06 10:17:39 [matchbox] [Query debug_companies_house] Using debug_companies_house
2025-06-06 10:18:36 [matchbox] [Query debug_companies_house] Finished
2025-06-06 10:19:23 [matchbox] [Model debug_naive_companies_house] Retrieving metadata
```

Np if you'd rather just get this in.

src/matchbox/client/dags.py
leo-mazzone: I made this unnecessary by using a memory SQLIte DB instead

PR #198, opened 2025-06-18
Propagate KeyboardInterrupts

sophie-daintta:

Not yet working as intended, but this draft PR provides a template of how server-side cancellation might work following client side KeyboardInterrupt. 

Issues: Query doesn’t appear to complete, although it seems to get stuck and hang either on second invocation or after cancellation. Cancelling the query manually in Postgres doesn’t solve the issue. 

## 🔗 Relevant links

Borrows from this blog. 
https://fastapiexpert.com/blog/2024/06/06/understanding-client-disconnection-in-fastapi/#check-client-disconnection 

## ✅ Checklist:

- [ ] All new code is tested
    - [ ] Tutorials
    - [ ] Developer docs

### DIFF
--- a/src/matchbox/server/api/main.py
+++ b/src/matchbox/server/api/main.py
@@ -1,14 +1,16 @@
 """API routes for the Matchbox server."""
 
 from importlib.metadata import version
-from typing import Annotated
+from typing import Annotated, Any, Awaitable
 
+import anyio
 from fastapi import (
     BackgroundTasks,
     Depends,
     FastAPI,
     HTTPException,
     Query,
+    Request,
     UploadFile,
     status,
 )
@@ -34,6 +36,7 @@ from matchbox.common.exceptions import (
     MatchboxSourceNotFoundError,
 )
 from matchbox.common.graph import ResolutionGraph
+from matchbox.common.logging import logger
 from matchbox.common.sources import Match
 from matchbox.server.api.arrow import table_to_s3
 from matchbox.server.api.cache import process_upload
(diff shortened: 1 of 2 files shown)


will-langdale:
Now I remember where I've seen AnyIO mentioned: it's part of FastAPI and Starlette.

I think my concern with this approach so far is that it puts async at the wrong level. Ideally, if the query endpoint is going to become explicitly asynchronous:

* We add an [asynchronous engine](https://docs.sqlalchemy.org/en/20/orm/extensions/asyncio.html) to PostgreSQL
* `backend.query()` also becomes asynchronous by using it with this engine

I think this is especially important because... what are you cancelling otherwise? `backend.query()` is blocking and the disconnect event won't process until it isn't. Nothing really gets cancelled. This leads to a broader challenge: 

> [!IMPORTANT]
> Is this ticket blocked by the implementation of an asynchronous backend?

Regardless, ideated a version based on this idea with AI:

```python
async def disconnected(request: Request) -> None:
    """Cancel task on client KeyboardInterrupt."""
    while True:
        message = await request.receive()
        if message["type"] == "http.disconnect":
            logger.info("Query disrupted by client disconnection.")
            break

@app.get("/query", responses={404: {"model": NotFoundError}})
async def query(
    request: Request,
    backend: BackendDependency,
    source: SourceResolutionName,
    resolution: ResolutionName | None = None,
    threshold: int | None = None,
    limit: int | None = None,
) -> ParquetResponse:
    """Query Matchbox for matches based on a source resolution name."""
    
    try:
        async with anyio.create_task_group() as tg:
            # Monitor for client disconnect (keyboard interrupt)
            tg.start_soon(disconnected, request)
            
            # Run the query directly
            res = await backend.query(
                source=source,
                resolution=resolution,
                threshold=threshold,
                limit=limit,
            )
            
    except MatchboxResolutionNotFoundError as e:
        raise HTTPException(
            status_code=404,
            detail=NotFoundError(
                details=str(e), entity=BackendRetrievableType.RESOLUTION
            ).model_dump(),
        ) from e
    except MatchboxSourceNotFoundError as e:
        raise HTTPException(
            status_code=404,
            detail=NotFoundError(
                details=str(e), entity=BackendRetrievableType.SOURCE
            ).model_dump(),
        ) from e
    except anyio.get_cancelled_exc_class():
        logger.info("Query cancelled due to client keyboard interrupt")
        raise HTTPException(status_code=499, detail="Client disconnected")

    buffer = table_to_buffer(res)
    return ParquetResponse(buffer.getvalue())
```

PR #202, opened 2025-06-23
Improve visualisation of the model subgraph

will-langdale:
The model visualisation function pretty consistantly produces things like this:

image

Awkward 🐛 

This PR amends it to produce things like this:

image

Beautiful 🦋 

* Changes to the visualisation function to make it plot nicer, and added a way to filter to only DAGs with a named resolution
* Fixed a bug where the fixtures for graph visualisations had a cycle, and had the parent/child relationship the wrong way around. Look at `_create_closure_entries()`: new resolutions are children, so sources should always be on the `parent` side of the relationship

## 👀 Guidance to review

Note that it wraps at two columns but I haven't had enough DAGs to test that.

## 🤖 AI declaration

Heavy use, all read and reviewed.

## 🔗 Relevant links

None.

## ✅ Checklist:

- [x] All new code is tested
    - [ ] Tutorials
    - [ ] Developer docs

### DIFF
--- a/src/matchbox/client/visualisation.py
+++ b/src/matchbox/client/visualisation.py
@@ -6,24 +6,52 @@ from matplotlib.figure import Figure
 from rustworkx.visualization import mpl_draw
 
 from matchbox.client._handler import get_resolution_graph
+from matchbox.common.dtos import ResolutionName
 from matchbox.common.graph import ResolutionNodeType
 
 
-def draw_resolution_graph() -> Figure:
-    """Draws the resolution graph layer by layer, component by component."""
+def draw_resolution_graph(contains: ResolutionName | None = None) -> Figure:
+    """Draws the resolution graph layer by layer, component by component.
+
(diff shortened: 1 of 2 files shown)


src/matchbox/client/visualisation.py
```diff
@@ -9,33 +10,86 @@
...
-            G,
-            pos={v: [0, i / 2] for v, i in enumerate(sources.values())},
-            fixed=set(sources.values()),
-        ),
-        node_color=colours,
-        with_labels=True,
-        labels=lambda node: node["name"],
-        font_size=8,
-    )
+    components = rx.weakly_connected_components(G)
```
sophie-daintta, line 18: Optional suggestion but would it be worth adding an option to filter components? ie. contains node = "naive_companies_house" or something similar?

sophie-daintta (commented):
One optional comment otherwise looks good to me.

sophie-daintta (approved):

PR #209, opened 2025-06-30
Remove auto-commit in large ingest

leo-mazzone:
Currently, our approach to dealing with atomicity of transactions, roll-backs etc. is fairly inconsistent. We need a proper review. For example, large_ingest doesn’t allow to insert multiple tables successfully or roll back.

* For both source indexing and results insertion:
  * Rollback on any kind of exception
  * Always re-raise the exception
  * Commit all new results, or nothing at all
* To enable this, now the function to ingest takes an ADBC connection managed by the caller
* Changed `MBDB.get_adbc_connection()` to return the pool proxy instead of the real ADBC connection
* Removed upserting logic, which wasn't used anywhere

## 👀 Guidance to review

* Currently, a failure in updating the cache, does not roll back the inserted tables, I suspect it's better this way.
* I've decided not to test the exception-rollback flows because the code is very simple.

## 🤖 AI declaration
None.

## 🔗 Relevant links
None.

## ✅ Checklist:

- [x] All new code is tested
    - [ ] Tutorials
    - [ ] Developer docs


src/matchbox/server/postgresql/utils/db.py
```diff
@@ -247,81 +253,83 @@ def large_ingest(
...
     if not update_columns and not upsert_keys:
         try:
             _copy_to_table(
                 table_name=table.name,
                 schema_name=table.schema,
+                connection=adbc_connection,
                 data=data,
                 max_chunksize=max_chunksize,
             )
+            return
```
will-langdale, line 266: Can we type hint the return type and tighten this up?

src/matchbox/server/postgresql/utils/db.py
```diff
@@ -247,81 +253,83 @@ def large_ingest(
...
+            )
+            upsert_stmt = insert_stmt.on_conflict_do_update(
+                index_elements=upsert_keys,
+                set_={c: getattr(insert_stmt.excluded, c) for c in update_columns},
+            )
 
-                session.execute(upsert_stmt)
-                session.commit()
+            session.execute(upsert_stmt)
+            session.commit()
```
will-langdale, line 320: Does this mean that:

* In regular mode, the transaction isn't committed
* In upsert mode, the transaction _is_ committed

From the dev's point of view, is this confusing? Should it be two functions?

src/matchbox/server/postgresql/utils/db.py
leo-mazzone: Done

src/matchbox/server/postgresql/utils/db.py
```diff
@@ -205,61 +208,112 @@ def _copy_to_table(
...
+                temp_table.drop(cleanup_session.bind, checkfirst=True)
+
+    def commit(self) -> None:
+        """Commit result of previous ingestions."""
+        if self._adbc_connection:
+            self._adbc_connection.commit()
+        if self._sqlalchemy_session:
+            self._sqlalchemy_session.commit()
+
+    def write(
```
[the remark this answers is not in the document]
leo-mazzone: I've removed upserting completely, but I've linked the permalink of my implementation in a new ticket

src/matchbox/server/postgresql/utils/db.py
```diff
@@ -205,61 +208,112 @@ def _copy_to_table(
...
+                    data=data,
+                    max_chunksize=max_chunksize,
+                )
+                return
+            except ADBCProgrammingError as e:
+                raise MatchboxDatabaseWriteError from e
 
-    # Upsert mode (slower)
-    else:
+        # -- UPSERT MODE (slower) --
```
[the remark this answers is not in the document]
leo-mazzone: Removed

src/matchbox/server/postgresql/utils/db.py
```diff
@@ -205,123 +207,50 @@ def _copy_to_table(
...
+            mode="append",
+            db_schema_name=schema_name,
+        )
 
 
-def large_ingest(
+def large_append(
     data: pa.Table,
     table_class: DeclarativeMeta,
+    adbc_connection: PoolProxiedConnection,
```
will-langdale, line 222: Suggest `connection` and check its type within the function so it matches `_copy_to_table()` and `sql_to_df()`, as it no longer needs differentiating.

PR #215, opened 2025-07-07
Feat/jwt authorisation

sophie-daintta:

This PR updates matchbox from basic authorisation with a shared matchbox api key to JSON Web Token (JWT) based authorisation of write endpoints based on EdDSA certificates. This may be considered as a partial solution to authorisation, as we currently don’t include any server-side user validation. Within the Data-Workspace architecture, JWT tokens are issued to select users via the MB__CLIENT__JWT environment variable. This PR includes:
 - Client-side JWT generation with 24 hour expiry.
 - (Optional) Server-side JWT validation on write endpoints.
 - Environment variable updates.
 - Additional testing for JWT expiry.

## 👀 Guidance to review

This solution could potentially be further developed within the FastAPI server to a full implementation of OAuth2. Alternatively, JWT authorisation could be moved away from FastAPI in the matchbox repository to a reverse proxy. We currently do not directly integrate with SSO for DBT beyond DW login. 

## ✅ Checklist:

- [x] All new code is tested


environments/sample.env
```diff
@@ -24,6 +25,7 @@ MB__SERVER__POSTGRES__ALEMBIC_CONFIG=
 
 MB__CLIENT__API_ROOT=
 MB__CLIENT__TIMEOUT=
-MB__CLIENT__API_KEY=
+MB__CLIENT__PRIVATE_KEY=
 MB__CLIENT__RETRY_DELAY=
 MB__CLIENT__DEFAULT_WAREHOUSE=
+
```
leo-mazzone, line 31: It might be time to create a sample client and sample server, i.e. split them, and also add the client user and JWT?

src/matchbox/server/api/dependencies.py
```diff
@@ -110,26 +116,62 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
 MetadataStoreDependency = Annotated[MetadataStore, Depends(metadata_store)]
 
 
-def validate_api_key(
-    settings: SettingsDependency, api_key: str = Security(API_KEY_HEADER)
+def b64_decode(b64_bytes):
+    """B64 decode."""
+    return urlsafe_b64decode(b64_bytes + (b"=" * ((4 - len(b64_bytes) % 4) % 4)))
```
leo-mazzone, line 121: Could you demystify this with a comment please

src/matchbox/client/authorisation.py
```diff
@@ -0,0 +1,60 @@
...
+import time
+from base64 import urlsafe_b64encode
+
+from cryptography.hazmat.primitives import serialization
+from cryptography.hazmat.primitives.asymmetric.ed25519 import Ed25519PrivateKey
+from cryptography.hazmat.primitives.serialization import load_pem_private_key
+
+from matchbox.client._settings import settings
+
+EXPIRY_AFTER_X_HOURS = 24
```
leo-mazzone, line 13: I'm confused, should this not be a server-side setting?

src/matchbox/server/api/dependencies.py
leo-mazzone: May I suggest

```
def b64_decode(b64_bytes):
    """Add padding and decode b64 bytes."""    
    remainder = len(b64_bytes) % 4
    if remainder:
        b64_bytes += b'=' * (4 - remainder)
    return base64.urlsafe_b64decode(b64_bytes)

```

src/matchbox/client/authorisation.py
sophie-daintta: More detail added in a doc string at the top of the authorisation file: 

"NOTE: The current version of this code is not structured for security as a standalone
Matchbox service but intended for providing authorisation within a pre-existing secure
platform such as Data Workspace. Client tokens should be generated within a secure
environment with private key access, from which they can be assigned to individual
users if required."

leo-mazzone (approved):

PR #221, opened 2025-07-20
feature: add security http headers

aidanrussell:

Security headers were requested to be added from the pentest review. This was considered a low severity. The headers do the following:
- responses cannot be cached
- https is enforced; any http requests are redirected to https
- the content-type must be stated, it will not be guessed if this field is missing (which is the default behaviour) - NOTE: is content-type always stated in the response? I was unsure about this.
- you can't frame the response

## 👀 Guidance to review

None.

## 🤖 AI declaration

None.

## 🔗 Relevant links

The instructions are in the jira ticket: https://uktrade.atlassian.net/browse/MTCHBX-428?atlOrigin=eyJpIjoiMDdiZDk3MGQ4YzkyNGJiNzlmZWUxMTg1MGM1Y2NjMjkiLCJwIjoiaiJ9

I found this resource useful as documentation on the effect of the headers: https://cheatsheetseries.owasp.org/cheatsheets/REST_Security_Cheat_Sheet.html

## ✅ Checklist:

- [ ] All new code is tested
    - [ ] Tutorials
    - [ ] Developer docs


aidanrussell:
Just to say @leo-mazzone that I tested this and it's behaving as expected - the headers are inserted into each response. I also checked that content-type was already being returned automatically by all the api calls in main.py and indeed that was the case so no to alter that (it is now required but was already in place).

PR #223, opened 2025-07-24
Add pooling to RelationalDBLocation

sophie-daintta:

Add connection pooling to RelationalDBLocation.

## 🤖 AI declaration

AI was used to, this code has been run in a development environment to test it. 

## 🔗 Relevant links

* [Connection pooling in SQLAlchemy engines](https://docs.sqlalchemy.org/en/20/core/engines.html#pooling)

## ✅ Checklist:

- [x] All new code is tested

### DIFF
--- a/src/matchbox/common/sources.py
+++ b/src/matchbox/common/sources.py
@@ -177,14 +177,6 @@ class RelationalDBLocation(Location):
         ),
     )
 
-    def __del__(self):
-        """Ensure connections are cleaned up when object is garbage collected."""
-        try:
-            if self.credentials:
-                self.credentials.dispose()
-        except Exception:
-            pass
-
     @field_validator("uri", mode="after")
     @classmethod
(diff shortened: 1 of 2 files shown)


src/matchbox/common/sources.py
```diff
@@ -176,12 +177,32 @@ class RelationalDBLocation(Location):
         ),
     )
 
+    def __del__(self):
+        """Ensure connections are cleaned up when object is garbage collected."""
+        try:
+            if self.credentials:
+                self.credentials.dispose()
+        except Exception:
+            pass
```
will-langdale, line 186: I've thought about this and imo we should remove it. The engine is shared between objects, and the user might want to do other things with it. Disposing of it overreaches our responsibility. Our other changes should be enough to use the pool properly.

will-langdale (commented):
Changed my mind about deletion.

PR #228, opened 2025-07-29
Use location names instead of URIs

leo-mazzone:
Our production set-up challenged a core assumption of the location system: one location has exactly one URI (and vice versa), by which it's defined. In reality:

* One location can have multiple URIs (as it happened to us in production across different environments)
* Multiple URIs (e.g. `sqlite:///:memory:`) can map to different locations

* Use location names, instead of URIs, to identify locations on the server, and for client to filter sources by location
* No validation is performed of the credentials added to a location. It's up to the user to ensure they set up the right credentials for a location. This means they could accidentally re-index a source from the wrong warehouse, or try to query a source and be unable to do so
* Allow using different credentials for different locations when downloading samples from server

### Normalisation
I didn't create a separate table for locations, in the spirit of "smallest, simplest solution to the problem". I'm not 100% sure this is right. One one hand, if we want to set rules on the creation of locations e.g. only for admins via CLI, that's when we should handle this, and at that point we might add a description with e.g. instructions / snippets for that location. On the other hand, it might mean a slightly more complex migration then to move existing data to a new table.

### Credentials
Now that locations don't have URIs, I'm not sure that "credentials" is the right name anymore for what we're passing. We're actually passing the engine that tells it where to look. Unfortunately "engine" might not be the right word either, depending on the type of location. Arango uses the words "clients" and "db" for different layers:

```python
from arango import ArangoClient
ArangoClient(hosts="http://localhost:8529").db("_system", username="root", password="passwd")
```

And S3 uses "client":

```python
boto3.client(
    's3',
    aws_access_key_id=ACCESS_KEY,
    aws_secret_access_key=SECRET_KEY,
    aws_session_token=SESSION_TOKEN
)
```

So maybe "client" is the right word. Thoughts?

## 🤖 AI declaration

AI only used to suggest `server_default` approach in migration.

## 🔗 Relevant links

* 

## ✅ Checklist:

- [ ] All new code is tested
    - [ ] Tutorials
    - [ ] Developer docs

### DIFF
--- a/src/matchbox/common/sources.py
+++ b/src/matchbox/common/sources.py
@@ -22,7 +22,6 @@ from typing import (
 import polars as pl
 from pyarrow import Table as ArrowTable
 from pydantic import (
-    AnyUrl,
     BaseModel,
     ConfigDict,
     Field,
@@ -35,7 +34,6 @@ from sqlalchemy.exc import OperationalError
 from matchbox.common.db import (
     QueryReturnType,
     ReturnTypeStr,
-    clean_uri,
     sql_to_df,
     validate_sql_for_data_extraction,
 )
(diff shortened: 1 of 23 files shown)


src/matchbox/common/sources.py
```diff
@@ -79,19 +77,19 @@ class Location(ABC, BaseModel):
     """A location for a data source."""
 
     type: LocationTypeStr
-    uri: AnyUrl
+    name: str
     credentials: Any | None = Field(exclude=True, default=None)
```
will-langdale, line 81: It's `client`, I feel pretty confident.

will-langdale (approved):
Looks good!

PR #229, opened 2025-07-29
Add tutorial on matchbox evaluation features

leo-mazzone:

* Add tutorial explaining evaluation features

## 👀 Guidance to review

* I'm not explaining how to actually launch the evaluation UI for now, because it's borderline impossible

## 🤖 AI declaration

I fed it my initial draft, and it made it much more concise and readable (naturally I reviewed and corrected the new version)

## ✅ Checklist:

    - [x] Tutorials

### DIFF
--- a/docs/client/evaluation.md
+++ b/docs/client/evaluation.md
@@ -1,3 +1,5 @@
+# Evaluate model performance
+
 After running models to deduplicate and link your data, you’ll likely want to check how well they’re performing. Matchbox helps with this by using **precision** and **recall**:
 
 * **Precision**: How many of the matches your model made are actually correct?
@@ -5,8 +7,8 @@ After running models to deduplicate and link your data, you’ll likely want to 
 
 To calculate these, we need a **ground truth** - a set of correct matches created by people. This is called **validation data**.
 
+## Creating validation data
 
-## Creating Validation Data
 Matchbox provides a simple UI to help you create this validation data. Here’s how it works:
 
 1. **Launch the UI** and choose a model resolution to sample clusters from.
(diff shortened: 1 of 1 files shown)


docs/client/evaluation.md
will-langdale, line 1: High level comments go here.

1. You have some double line breaks
2. I personally like a line break after a heading
3. I think elsewhere we've used sentence case for headings -- I think it's easier to read

docs/client/evaluation.md
```diff
@@ -0,0 +1,129 @@
...
+p, r = eval_data.precision_recall(results, threshold=0.5)
+```
+
+!!! tip "Deterministic models"
+    Some types of model (like the `NaiveDeduper` used in the example) only output 1s for the matches they make, hence **threshold truth tuning doesn't apply**:
+
+    * You won't get a precision-recall curve, but a single point at threshold 1.
+    * The precision and recall scores will be the same at all thresholds.
+
+    On the other hand, probabilistic models (like `SplinkLinker`), can output **any integer between 0 and 1**.
```
will-langdale, line 107: Any float.

Wonder if this section should say `1.0` and `0.0` (on 102 as well).

docs/client/evaluation.md
```diff
@@ -0,0 +1,129 @@
...
+    "resolution_name_two",
+    "resolution_name_three"
+]
+comparison = compare_models(models_to_compare)
+
+for model in models_to_compare:
+    p, r = comparison[model]
+    print(f"Model {model} has precision: {p} and recall: {r}")
+```
+!!! tip "Model records overlap"
```
will-langdale, line 128: line break

docs/client/evaluation.md
leo-mazzone: Done

docs/client/evaluation.md
leo-mazzone: Done

PR #232, opened 2025-07-30
feature: logging platform related code is separated out

aidanrussell:

Remove all datadog related code to be stored in a private repository. Publish now-cleaned image using Github packages as an open-source docker container.

## 👀 Guidance to review

I would encourage having a think if I have missed anything. Also part of the intent here is to ensure we are aligned to understand what this task involves so do have a think if it meets the expectations you had in advance. 

NOTE: This MR shouldn't actually be merged until the second repository solution is completed to allow continuity of logging.

## 🤖 AI declaration

None.

## 🔗 Relevant links

Related MR in matchbox-deploy: https://github.com/uktrade/matchbox-deploy/pull/9

## ✅ Checklist:

- [ ] All new code is tested
    - [ ] Tutorials
    - [ ] Developer docs


aidanrussell:
Some discussion points on this one:
- I slightly rearranged pyproject.toml as I found it a bit confusing to have some "groups" and some "optional dependencies" so they are now all "groups"
- I also reordered some packages to be in alphabetical order in pyproject.toml (i.e. the only significant change to this file if you check carefully is removing ddtrace package)
- I had an insurmountable issue regarding the sha256 value for sentry-sdk in uv.lock being mistakenly taken as a sentry key value by Trufflehog. Much as I might be a bit nervous to disregard Trufflehog, I was finally only able to overcome this by excluding uv.lock from Trufflehog - I'm confident this is the correct thing to do anyway given it is an exclusively auto-generated file.

.github/workflows/deploy-docker-image-to-packages.yml
```diff
@@ -0,0 +1,51 @@
+name: Create and Publish the Matchbox Open-Source Docker Image
+
+on:
+  push:
+    branches:
+      - 'main'
```
will-langdale, line 6: Bit of a tension.

On release makes most sense for prod, but staging deploys on pushes to main.

Can we tag staging with .dev or whatever? And should we be clearing these up? I worry if we do it on pushes to main we'll flood the thing.

PR #234, opened 2025-07-31
Removed all cleaning logic from Matchbox

will-langdale:
Cleaning is causing me more problems than it solves. I reached my limit when I wanted to aggregate to arrays then pick the first item: Matchbox was designed in a way that would make the steps to implement this ridiculous in their complexity.

I believe that cleaning is not Matchbox's concern. If you want a library of cleaning functions, import it from elsewhere.

I therefore replace all cleaning with a simple `StepInput.cleaning_dict` which will run DuckDB SQL against the step's input.

* All cleaning logic gone
* Adds `StepInput.cleaning_dict` with small unit tests
* A bit of extra editing in `link-data.md`
    * Demonstrate source name qualification
    * Removed "conditional matching" section as the deterministic linker now works differently and it's quite a specific thing in a more general tutorial

## 👀 Guidance to review

Nothing special: it's mostly deletion.

My biggest worry is I'll have missed something in the documentation, but I searched for "clean" and "process" and it looks like it's been, well, cleaned.

## 🤖 AI declaration

Heavy use in quickly refactoring lots of bitty changes, but all reviewed.

## 🔗 Relevant links

* [Writing pipelines Confluence](https://uktrade.atlassian.net/wiki/spaces/CDL/pages/4891934779/Writing+pipelines)

## ✅ Checklist:

- [x] All new code is tested
    - [x] Tutorials
    - [ ] Developer docs


docs/client/link-data.md
```diff
@@ -447,34 +424,44 @@ You can link across multiple sources in a single step:
         left=StepInput(
             prev_node=dedupe_companies,
             select={companies_house: ["company_name", "postcode"]},
-            cleaners=ch_clean_simple,
+            cleaning_dict={
+                "company_name": f"lower({companies_house.f('company_name')})",
+                "postcode": companies_house.f("postcode"),
```
leo-mazzone, line 429: this is unnecessary, and could you also explain in the docs the point about new fields vs old fields vs overwritten fields?

src/matchbox/client/helpers/clean.py
```diff
@@ -0,0 +1,92 @@
...
+    def _composed_transform(column: str) -> str:
+        result = column
+        for func in functions:
+            result = func(result)
+        return result
+
+    return _composed_transform
+
+
+def vectorise(func: Callable[[str], str]) -> Callable[[str], str]:
```
leo-mazzone, line 54: Similar to comment above, I don't think we want to enforce the inputs and outputs of the callable. Could we instead take as input to `vectorise` the SQL itself? That way it's much more flexible, it just applies vectorially the SQL output by another function or combination thereof

docs/client/link-data.md
will-langdale: All done.

src/matchbox/client/helpers/clean.py
```diff
@@ -0,0 +1,92 @@
+"""Utility functions for composing SQL cleaning expressions."""
+
+from typing import Callable
+
+
+def compose(*functions: Callable[[str], str]) -> Callable[[str], str]:
```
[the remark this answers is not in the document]
will-langdale: Agreed to remove -- should add because we've found it useful, not because we think we should.

leo-mazzone (approved):
The day we remove 1500 lines of code is a good day

leo-mazzone (changes requested):
if the flow is

`analytical interface -> DAG interface`

it might be useful to move the whole logic of `StepInput.clean()` to a function `helpers.selector.clean()` so that users can easily port whatever cleaning logic they're working on to a DAG. Then `StepInput.run()` could just apply this more generic function

PR #235, opened 2025-08-04
Improve concurrency model

leo-mazzone:

* Turn all routes interacting with S3 or Postgres to use sync functions
  * Also affects some functions on which those routes depend
* 

## 👀 Guidance to review

* I've kept the health-check endpoints and FastAPI dependencies async, see relevant links for details
  * Key quote is "If your application (somehow) doesn't have to communicate with anything else and wait for it to respond, use async def, even if you don't need to use await inside."

## 🔗 Relevant links

* FastAPI docs: [using files in synchronous context](https://fastapi.tiangolo.com/tutorial/request-files/#uploadfile)
* FastAPI docs: https://fastapi.tiangolo.com/async/

## ✅ Checklist:

- [ ] All new code is tested
    - [ ] Tutorials
    - [ ] Developer docs

### DIFF
--- a/src/matchbox/server/api/arrow.py
+++ b/src/matchbox/server/api/arrow.py
@@ -16,7 +16,7 @@ else:
     S3Client = Any
 
 
-async def table_to_s3(
+def table_to_s3(
     client: S3Client,
     bucket: str,
     key: str,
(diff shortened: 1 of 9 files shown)


will-langdale (changes requested):
With one change, happy.

will-langdale (approved):
You are smart, I am not. Approved.

PR #241, opened 2025-08-08
Feature: publish open source matchbox image

aidanrussell:

Create a new version of the matchbox Dockerfile that is clean of any private configurations specific to the existing deployment (in this case, datadog-specific configuration but intended as a general principle going forward). Publish that as an open-source image downloadable with `docker pull` command. Simultaneously create solution for the removed configuration to be deployed using matchbox-deploy private repository (see related MR in that repository). 

## 👀 Guidance to review

None.

## 🤖 AI declaration

None.

## 🔗 Relevant links

Related MR in matchbox-deploy: https://github.com/uktrade/matchbox-deploy/pull/9
Related MR in data-workspace: tbc

## ✅ Checklist:

- [ ] All new code is tested
    - [ ] Tutorials
    - [ ] Developer docs


.github/workflows/deploy-docker-image-to-packages.yml
```diff
@@ -0,0 +1,51 @@
+name: Create and Publish the Matchbox Open-Source Docker Image
+
+on:
+  push:
+    branches:
+      - 'main'
```
will-langdale, line 6: Where did we come down pushes to main vs releases? This config will mean we're putting unstable versions on the latest tag.

If possible, maybe the answer is:

* Release creates tag `vx.x.x:latest`
* Merge to main creates tag `v.x.x.x+1dev`, overwriting the previous image if there is one

What do you think @leo-mazzone

.github/workflows/deploy-docker-image-to-packages.yml
```diff
@@ -0,0 +1,51 @@
...
+jobs:
+  build-and-push-image:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+      attestations: write
+      id-token: write
+    steps:
+      - name: Checkout repository
```
will-langdale, line 21: Can we do linebreaks between name specs? Makes it much easier to read.

.github/workflows/deploy-docker-image-to-packages.yml
aidanrussell: I have done this, although I wonder if keeping only the release version may be sufficient

will-langdale: I'm seeing a lot of the images on there are named like `matchbox:0.1.dev1-g45e3d4f`. I'd expect to see:

* For release, `matchbox:0.4.10` (aka `:latest`)
* For merge to main `matchbox:0.4.11-dev` (where this image is overwritten with each merge to main)

Actually, I just came across [docker-metadata-action](https://github.com/marketplace/actions/docker-metadata-action) and with a little tweak to the plan, I think we can simplify those two actions significantly.

The below action YAML will create:

* On release, `matchbox:0.4.10` (aka `:latest`)
* On merge to main, `main-development`

The one action fires on both triggers, and the metadata action just slaps the appropriate tags on.

On reflection I think this actually much neater. We only ever use and test one development version in staging, so let's just have one in the registry rather than having hundreds hanging around. The constraint of the action has led to, I think, a more appropriately concise solution.

This is AI code and needs proper reviewing and testing.

```yaml
name: Publish Docker image to GitHub Container Registry

on:
  push:
    branches: [main]
  release:
    types: [published]

env:
  REGISTRY: ghcr.io
  IMAGE_NAME: ${{ github.repository }}

jobs:
  build-and-push:
    runs-on: ubuntu-latest
    
    permissions:
      contents: read
      packages: write

    steps:
      - name: Checkout
        uses: actions/checkout@v4

      - name: Login to GitHub Container Registry
        uses: docker/login-action@v3
        with:
          registry: ${{ env.REGISTRY }}
          username: ${{ github.actor }}
          password: ${{ secrets.GITHUB_TOKEN }}

      - name: Install uv
        uses: astral-sh/setup-uv@v6

      - name: Extract version for build arg
        run: echo "MB_VERSION=$(uv run --frozen python -m setuptools_scm)" >> $GITHUB_ENV

      - name: Extract metadata
        id: meta
        uses: docker/metadata-action@v5
        with:
          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
          tags: |
            type=ref,event=branch,suffix=-development
            type=semver,pattern={{version}}
            type=raw,value=latest,enable=${{ github.event_name == 'release' }}

      - name: Build and push
        uses: docker/build-push-action@v6
        with:
          context: .
          file: src/matchbox/server/Dockerfile
          push: true
          build-args: MB_VERSION=${{ env.MB_VERSION }}
          tags: ${{ steps.meta.outputs.tags }}
          labels: ${{ steps.meta.outputs.labels }}
```

.github/workflows/deploy-docker-image-to-packages.yml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Run uv to extract the version and set as an environment variable
+        run:  echo "MB_VERSION=$(uv run --frozen python -m setuptools_scm)" >> $GITHUB_ENV
+
+      - name: Extract tag metadata for Docker
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          tags: |
+            type=raw,value=${{ env.MB_VERSION }}
```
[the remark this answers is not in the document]
aidanrussell: I'm not following you here - hundreds of releases for every push to main? Let's be clear in the language first of all so the two actions are merge to main and release. Both of these are manually created. Each one creates a new MB_VERSION. I think this is all one to one in that for either a merge or a release, you'll get one new MB_VERSION and so one new image. That image can then also be tagged either "latest" (for releases) or "development" (for merges). I think what is here already is correct?

PR #254, opened 2025-08-28
feature: add curl as an aid to deployment healthchecks

aidanrussell:

Add curl to the Docker image. The reason is this is useful for health checks when deployed. It could go in matchbox-deploy but my view is curl is sufficiently widely used for this purpose to the point that any open-source user would likely expect it to be there. 

If you agree and this is merged then I'll remove it from the version of matchbox-deploy I've got currently which goes together with adding these health checks on an MR in data-workspace (i.e. we want to make sure curl is there, either from matchbox or from matchbox-deploy, before that change would be applied to prod/staging or the health checks fail!).

## 🤖 AI declaration

None.

## ✅ Checklist:

- [ ] All new code is tested
    - [ ] Tutorials
    - [ ] Developer docs

### DIFF
--- a/src/matchbox/server/Dockerfile
+++ b/src/matchbox/server/Dockerfile
@@ -1,7 +1,7 @@
 FROM python:3.13-slim AS base
 
-# Install git (required for setuptools_scm)
-RUN apt-get update && apt-get install -y git && rm -rf /var/lib/apt/lists/*
+# Install git (required for setuptools_scm) and curl (as an aid for healthchecks)
+RUN apt-get update && apt-get install -y git curl && rm -rf /var/lib/apt/lists/*
 
 COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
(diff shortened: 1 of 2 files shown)


src/matchbox/server/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM python:3.13-slim AS base
 
-# Install git (required for setuptools_scm)
-RUN apt-get update && apt-get install -y git && rm -rf /var/lib/apt/lists/*
+# Install git (required for setuptools_scm) and curl (as an aid for healthchecks)
```
will-langdale, line 3:
```suggestion
# Install git (required for setuptools_scm) and curl (as an aid for healthchecks)
RUN apt-get update && \
    apt-get install -y git curl && \
    rm -rf /var/lib/apt/lists/*
```

will-langdale (approved):
👍

PR #278, opened 2025-09-15
Hardened celery logic, adding retry and timeout

will-langdale:
Our uploads will sometimes run forever. This gives them a hard limit.

* Adds hard limit to upload time and extra try/catch/retry for celery tasks

## 👀 Guidance to review

Note none of it's tested beyond integration tests. It's all pretty core celery stuff and testing it is harder than it's worth.

Also I don't like the indented `try`/`except` but all the key ways of checking "has this happened more than the retry count" rely on undocumented attributes of the `Request` object. Even the way _Celery itself uses_ uses this attribute. I've chosen the route of crap code I hate, but it'll work.

## 🤖 AI declaration

Ideated with AI, implemented by hand.

## 🔗 Relevant links

* [Celery Task documentation](https://docs.celeryq.dev/en/stable/userguide/tasks.html#retrying)

## ✅ Checklist:

- [ ] All new code is tested
    - [ ] Tutorials
    - [ ] Developer docs

### DIFF
--- a/src/matchbox/server/uploads.py
+++ b/src/matchbox/server/uploads.py
@@ -9,10 +9,8 @@ from typing import TYPE_CHECKING, Any
 
 import pyarrow as pa
 import redis
-from celery import Celery
-from fastapi import (
-    UploadFile,
-)
+from celery import Celery, Task
+from fastapi import UploadFile
 from pyarrow import parquet as pq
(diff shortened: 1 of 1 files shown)


leo-mazzone (approved):
Looks good

PR #294, opened 2025-10-06
Fix prerelease MB version identification

leo-mazzone:
Previous attempt to get pre-release to publish dev version of package to PyPi failed.

* Define step output in pre-release YAML as an attempt to fix broken workflow

## 👀 Guidance to review

None

## 🤖 AI declaration

None

## 🔗 Relevant links

* [GitHub guide on passing output between steps](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/pass-job-outputs)

## ✅ Checklist:

- [ ] All new code is tested
    - [ ] Tutorials
    - [ ] Developer docs

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -9,6 +9,8 @@ env:
 jobs:
   build:
     runs-on: ubuntu-latest
+    outputs:
+      mb_version: ${{ steps.get_version.outputs.mb_version }}
     permissions:
       contents: write
       packages: write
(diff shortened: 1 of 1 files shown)


sergio-recrod (approved):
Checked:
- Necessary permissions given
- Command to return version works as intended and output is used where needed
- Needed steps match those in the release workflow

A few things to confirm:
- Step `Build and push Docker image` is different in prerelease workflow
- Step to create gh release has some dependencies in the release workflow, don't we need those in the prerelease?

Issue #318, opened 2025-10-23
Possible to deserialise invalid strings for extract transform logic


will-langdale:
Crash when reloading a source that contained extract transform logic that:

* Was a raw string
* Contained unescaped quotation marks
* Contained regex

I can't share the full SQL that caused it but here's the trace. Should be an easy thing to add to test coverage, then fix.

## Acceptance

Fix the bug

## Trace

```
---------------------------------------------------------------------------
ParseError                                Traceback (most recent call last)
File /opt/conda/envs/company-pipelines/lib/python3.12/site-packages/matchbox/client/locations.py:222, in RelationalDBLocation.validate_extract_transform(self, extract_transform)
    221 try:
--> [222](https://vscode-remote+vscode-002deaf4fd9a-002edata-002etrade-002egov-002euk.vscode-resource.vscode-cdn.invalid/opt/conda/envs/company-pipelines/lib/python3.12/site-packages/matchbox/client/locations.py:222)     expressions = sqlglot.parse(extract_transform, dialect=dialect)
    223 except ParseError as e:

File /opt/conda/envs/company-pipelines/lib/python3.12/site-packages/sqlglot/__init__.py:102, in parse(sql, read, dialect, **opts)
     90 """
     91 Parses the given SQL string into a collection of syntax trees, one per parsed SQL statement.
     92 
   (...)    100     The resulting syntax tree collection.
    101 """
--> [102](https://vscode-remote+vscode-002deaf4fd9a-002edata-002etrade-002egov-002euk.vscode-resource.vscode-cdn.invalid/opt/conda/envs/company-pipelines/lib/python3.12/site-packages/sqlglot/__init__.py:102) return Dialect.get_or_raise(read or dialect).parse(sql, **opts)

File /opt/conda/envs/company-pipelines/lib/python3.12/site-packages/sqlglot/dialects/dialect.py:1101, in Dialect.parse(self, sql, **opts)
   1100 def parse(self, sql: str, **opts) -> t.List[t.Optional[exp.Expression]]:
-> [1101](https://vscode-remote+vscode-002deaf4fd9a-002edata-002etrade-002egov-002euk.vscode-resource.vscode-cdn.invalid/opt/conda/envs/company-pipelines/lib/python3.12/site-packages/sqlglot/dialects/dialect.py:1101)     return self.parser(**opts).parse(self.tokenize(sql), sql)

[...]

PR #321, opened 2025-10-23
Fix query validation for loaded sources

sergio-recrod:
ETL validation was not working in instances where sources were loaded from the server. Because they did not have a client, the validation could not infer the SQL dialect to check. 

Fixes https://github.com/uktrade/matchbox/issues/318

We added a flag to Source to indicate whether the source ETL should be validated. This is set to False when loading a source from the server.

We assume that any source already uploaded to the server would have been validated when the DAG was first run.

## 👀 Guidance to review

N/A

## 🤖 AI declaration

N/A

## 🔗 Relevant links

N/A
## ✅ Checklist:

- [x] All new code is tested
    - [ ] Tutorials
    - [ ] Developer docs

### DIFF
--- a/src/matchbox/client/sources.py
+++ b/src/matchbox/client/sources.py
@@ -105,9 +105,6 @@ class Source:
                 perform query validation. It should be False when loading sources from
                 the server. Default True.
         """
-        # Only validate if client is present
-        # A client MUST be present if initialising source for the first time
-        # (e.g. dag.source)
         if validate_etl:
             location.validate_extract_transform(extract_transform)
 


src/matchbox/client/dags.py
```diff
@@ -110,6 +110,8 @@ def final_step(self) -> Source | Model:
 
     def source(self, *args, **kwargs) -> Source:
         """Create Source and add it to the DAG."""
+        if not kwargs["location"]:
```
leo-mazzone: and the client can't be None

PR #331, opened 2025-10-24
Adds e2e tests for methodologies, and fixes ensuing bugs

### LINKED ISSUE #287: Joining clusters to their source will suffer from PK clashes
See `Results.inspect_clusters()` and one of the comments in [PR34](https://github.com/uktrade/matchbox/pull/34).

In short, the source data’s primary key isn’t guaranteed to be unique – and indeed is likely not to be in tables with autoincrementing integer keys.

We need to qualify the source name to the dataframe before joining the source data and using it as a left/right_key/merge_col.

### LINKED ISSUE #320: SplinkLinker broken when more index fields exist than it joins on
`Query.run()` will select and pass on all fields by default, because it has to load the whole source into memory anyway. 

`Query.clean()` will pass on any fields not being cleaned.

This means there's no clear method to drop fields before they're passed to a model, and Splink requires both datasets to be identical.

Me and @leo-mazzone agree the behaviour of `Query.run()` should return all columns when nothing is specified.

## Acceptance

Fix the `SplinkLinker` when there's more index columns than it's matching on.

will-langdale:
The SplinkLinker was still running into problems in production. This deals with the root issue: cleaning was returning the key field.

Fixes #320, fixes #287 (or makes it irrelevant by removing the buggy feature)

* Creates e2e tests for methodologies using the scenario system
    * Which meant updating the configurator functions
* Removes keys from any `Query.run()` functions using cleaning, so you can still select them if you wish
* Removes unused `Results.inspect_*` functions. While I suspect they'll be useful, they're broken and untested, and should be added back as they're needed

## 👀 Guidance to review

None.

## 🤖 AI declaration

Used to ideate and explore the problem. Wrote all the tests, which were carefully reviewed.

## 🔗 Relevant links

None.

## ✅ Checklist:

- [x] All new code is tested
    - [ ] Tutorials
    - [ ] Developer docs

### DIFF
--- a/src/matchbox/client/queries.py
+++ b/src/matchbox/client/queries.py
@@ -296,13 +296,11 @@ class Query:
 def _clean(
     data: pl.DataFrame,
     cleaning_dict: dict[str, str] | None,
-    key_columns: list[str] | None = None,
 ) -> pl.DataFrame:
     """Clean data using DuckDB with the provided cleaning SQL.
 
     * ID is passed through automatically
     * If present, leaf_id is passed through automatically
-    * Key columns (specified in key_columns) are passed through automatically
     * Columns not mentioned in the cleaning_dict are dropped
     * Each key in cleaning_dict is an alias for a SQL expression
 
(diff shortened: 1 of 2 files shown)


leo-mazzone:
Fixes https://github.com/uktrade/matchbox/issues/287
Or rather, makes it irrelevant

Issue #333, opened 2025-10-25
Cannot update resolution in pending


leo-mazzone:
If local and remote resolution configs diverge, you get a value error, and no way of deleting the resolution.

## Acceptance
- Allow to load pending run, change a model or source, and update state on the server

Issue #298, opened 2025-10-09
Mutable runs don't clear down data properly


will-langdale:
This issue will be raised by #284 but we've made it out of scope. Spotted by @leo-mazzone 

When data is uploaded for a mutable run, it should clear down the previous data completely. Consider the situation where your run returns radically fewer probabilities: right now, only the new probabilities would be added, and the old ones would be kept.

## Acceptance

* Backend adapter tests that check that setting data (for both sources and modals) on mutable runs clears down data fully
* Backend adapter tests that check that setting data cannot be done for immutable runs
* The PostgresSQL adapter passes these tests

Issue #297, opened 2025-10-08
409 assumes deletion only


will-langdale:
The server raises 409s for a variety of situations: deletion not confirmed AND other invalid parameters.

The handler currently only processes the deletion confirmation 409.

Fix it.

PR #336, opened 2025-10-27
Simplify DAG run model

leo-mazzone:
Fixes https://github.com/uktrade/matchbox/issues/298
Fixes https://github.com/uktrade/matchbox/issues/333

* Renamed resolution hash to "fingerprint" to avoid confusion, and added it to the DTO for resolutions. This is now computed client-side before creating a new resolution, and is used to check divergence with resolution data on the server
* Create endpoint and backend method to update resolution metadata, which replaces truth setting / getting
* Simplified upload tracker, and added upload stage to `Resolution` on the ORM.
* Moved upload endpoints under router for collections, and collapsed endpoint to get upload ID and endpoint to upload data
* Improved node sync logic: error if node not run; then check fingerprint. If it diverges, delete resolution and start over. If it matches, patch resolution. Then check upload status: if it's ready, always upload data, otherwise skip.
* Remove `last_run` attribute on DAG nodes. Now, if a node is run either individually or through the DAG, it's always run. This simplifies internal state management ahead of further changes, and puts on the user the onus of deciding when to re-run things
* Simplified logic for drawing a DAG with status indicators

## 👀 Guidance to review
* **BREAKING CHANGES**: The hash field of a resolution (renamed to fingerprint) is now non-nullable. However, filling with a constant should take care of that. This will create a mis-match between fingerprint and data, but will allow database to migrate. I've also added the upload stage field to resolutions, but set it to "complete" for all past resolutions. 
* Increasingly I think we need to reduce duplication between client `Source` and `Model` with some inheritance. Considered out of scope for now
* The client-side translation between HTTP error code and exception is all a bit broken. Considered out of scope
* This logic is not resistant in any way to race conditions. Considered out of scope
* Forcing a `Resolution` to have a fingerprint has made tests more annoying to write. I've had to create a `.fake_run()` method on testkits, to assign their hashes / probabilities to the source / model, as if those had actually run
* I've deleted a couple of 404 API tests, hope you don't mind. I think we have been over-testing things that are too trivial
* I've removed the resolution type validation in the adapter because it wasn't useful and was being used inconsistently across adapter methods

## 🤖 AI declaration

Helped generate the Alembic migration, the new endpoint, and to re-apply source test refactor to model test.

## ✅ Checklist:

- [x] All new code is tested
    - [x] Tutorials
    - [ ] Developer docs


will-langdale:
> The client-side translation between HTTP error code and exception is all a bit broken. Considered out of scope

See #297 

> This logic is not resistant in any way to race conditions

Can you raise an issue please?

Finally, what's the impact of this PR on #298? Feels like a fix, perhaps?

src/matchbox/client/models/models.py
```diff
@@ -48,6 +51,24 @@ def add_model_class(ModelClass: type[Linker] | type[Deduper]) -> None:
         raise ValueError("The argument is not a proper subclass of Deduper or Linker.")
 
 
+def post_run(method: Callable[..., T]) -> Callable[..., T]:
```
will-langdale: If you're feeling fancy:

```python
from typing import Protocol, TypeVar, Callable

class HasResults(Protocol):
    """Protocol for objects that have a results attribute."""
    results: ArrowTable | None

S = TypeVar('S', bound=HasResults)
T = TypeVar('T')

def post_run(method: Callable[[S, ...], T]) -> Callable[[S, ...], T]:
```

src/matchbox/client/sources.py
```diff
@@ -38,6 +38,24 @@
 T = TypeVar("T")
 
 
+def post_run(method: Callable[..., T]) -> Callable[..., T]:
```
will-langdale: As with models, for fancy lads only.

```python
from typing import Protocol, TypeVar, Callable

class HasHashes(Protocol):
    """Protocol for objects that have a hashes attribute."""
    hashes: ArrowTable | None

S = TypeVar('S', bound=HasHashes)
T = TypeVar('T')

def post_run(method: Callable[[S, ...], T]) -> Callable[[S, ...], T]:
```

Potentially has the benefit of being a bit clearer that this isn't _quite_ replication, even though it feels like it.

src/matchbox/server/api/routers/collection.py
```diff
@@ -490,107 +577,163 @@ def delete_resolution(
...
+    client = backend.settings.datastore.get_client()
+    bucket = backend.settings.datastore.cache_bucket_name
+    key = f"{upload_id}.parquet"
 
     if resolution.resolution_type == ResolutionType.SOURCE:
-        upload_id = upload_tracker.add_source(path=resolution_path)
-        return upload_tracker.get(upload_id=upload_id).status
+        expected_schema = SCHEMA_INDEX
+    else:
+        expected_schema = SCHEMA_RESULTS
```
will-langdale: I'd do this all before changing the status -- this is all stuff that might error and we can save a headache but not changing the status yet, I think?

src/matchbox/server/postgresql/adapter.py
```diff
@@ -349,14 +359,88 @@ def create_resolution(  # noqa: D102
...
+                )
+
+            # Update top-level metadata
+            old_resolution.description = resolution.description
+            old_resolution.truth = resolution.truth
+
+            # Update config
+            if old_resolution.type == "source":
+                old_config: SourceConfigs = old_resolution.source_config
+                assert isinstance(new_config, CommonSourceConfig)
```
will-langdale: No asserts in production code.

src/matchbox/server/postgresql/adapter.py
```diff
@@ -349,14 +359,88 @@ def create_resolution(  # noqa: D102
...
+                            )
+                            for idx, field in enumerate(new_config.index_fields)
+                        ]
+                    )
+
+                    old_config.fields = new_fields
+
+            else:
+                old_config: ModelConfigs = old_resolution.model_config
+                assert isinstance(new_config, CommonModelConfig)
```
will-langdale: No asserts in production code. Might consider adding [S101](https://docs.astral.sh/ruff/rules/assert/#assert-s101). The whole of `S` feels like overkill right now.

test/client/test_eval.py
```diff
@@ -77,11 +77,21 @@ def test_get_samples(
...
         resolutions={
-            "foo": foo_testkit.source.to_resolution(),
-            "bar": bar_testkit.source.to_resolution(),
-            "baz": baz_testkit.source.to_resolution(),
-            "linker1": foo_bar.to_resolution(),
-            "linker2": foo_bar_baz.to_resolution(),
+            "foo": foo_testkit.fake_run().source.to_resolution(),
+            "bar": bar_testkit.fake_run().source.to_resolution(),
+            "baz": baz_testkit.fake_run().source.to_resolution(),
+            "linker1": Resolution(
```
will-langdale: Why not `ModelTestkit.fake_run()`?

src/matchbox/server/postgresql/adapter.py
leo-mazzone: Done

src/matchbox/server/postgresql/adapter.py
```diff
@@ -378,17 +462,34 @@ def delete_resolution(self, path: ResolutionPath, certain: bool) -> None:  # noq
 
     # Data insertion
 
+    def set_resolution_stage(self, path: ResolutionPath, stage: UploadStage) -> None:  # noqa: D102
+        self._check_writeable(path)
```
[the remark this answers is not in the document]
leo-mazzone: Nope

src/matchbox/server/postgresql/alembic/versions/59d47cc16734_make_resolution_fingerprint_non_nullable.py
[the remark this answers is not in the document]
leo-mazzone: Done

src/matchbox/server/api/main.py
```diff
@@ -140,11 +127,11 @@ async def deletion_not_confirmed_handler(
     """Handle deletion not confirmed errors."""
     # Extract resource name from request
     path_parts = request.url.path.split("/")
-    resource_name = path_parts[-1] if path_parts else "unknown"
+    resource_name = path_parts[-1] if path_parts else "Unknown"
```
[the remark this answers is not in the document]
leo-mazzone: We agreed to not do this for now

src/matchbox/server/api/routers/collection.py
leo-mazzone: I've moved it to right before background task

src/matchbox/server/api/routers/collection.py
```diff
@@ -490,107 +577,163 @@ def delete_resolution(
...
+            detail=ResourceOperationStatus(
+                success=False,
+                target=f"Resolution data {resolution_path}",
+                operation=CRUDOperation.CREATE,
+                details=f"Expected .parquet file, got {file.filename.split('.')[-1]}",
+            ),
+        )
+
+    # pyarrow validates Parquet magic numbers when loading file
+    try:
```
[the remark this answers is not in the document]
leo-mazzone: done

src/matchbox/server/api/routers/collection.py
```diff
@@ -490,107 +577,163 @@ def delete_resolution(
...
+                details="Not expecting upload for this resolution.",
+            ),
+        )
+    # Signal to clients that resolution is not ready to receive data
+    backend.set_resolution_stage(path=resolution_path, stage=UploadStage.PROCESSING)
+    resolution = backend.get_resolution(path=resolution_path)
+
+    upload_id = str(uuid.uuid4())
+
+    # Validate file
```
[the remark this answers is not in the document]
leo-mazzone: done

src/matchbox/server/api/routers/collection.py
```diff
@@ -490,107 +577,165 @@ def delete_resolution(
...
+                target=f"Resolution data {resolution_path}",
+                operation=CRUDOperation.CREATE,
+                details=f"Could not upload file to object storage: {str(e)}",
+            ),
+        ) from e
 
-    upload_id = upload_tracker.add_model(path=resolution_path)
-    return upload_tracker.get(upload_id=upload_id).status
+    # Start background processing
+    backend.set_resolution_stage(path=resolution_path, stage=UploadStage.PROCESSING)
```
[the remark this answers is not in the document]
leo-mazzone: I found it simpler to think about a state where only the background / celery task is responsible for resetting a resolution. I dislike implementing half-baked things because I think it gives us a false sense of security. We are not resistant to race conditions right now period and the S3 upload is only a small part of that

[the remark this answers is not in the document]
leo-mazzone: https://github.com/uktrade/matchbox/issues/346

will-langdale (approved):
Well done!

PR #379, opened 2025-12-02
Feat/profiling logs

sergio-recrod:
Added decorator for profiling and benchmarking. 

- Decorator added to logging that can be used in any function to log processing time.
- Added decorator to a set of relevant functions.

## 👀 Guidance to review

I've considered two use cases for this decorator:
- Decorate functions to get running time. The decorator can take a `kwarg` argument that will take a kwarg name of the function being evaluated and add it with its value to the logging message. For instance, if decorating `set_data`, we'd get:
```bash
[Profiling] `set_data` with path `companies_mock_benchmark/82/ch_deduped` took 4.063s
```
- Decorate methods within a class. I can see this happening for classes `Source` and `Model`. In this case, I think the most informative attribute to log is the node name. For instance, if decorating `Source.fetch`, we would get:
```bash
[Profiling] `fetch` in node `ch` took 0.000s
```

## 🤖 AI declaration

Some Q&A to figure out how to access class attributes to log them.

## ✅ Checklist:

- [ ] All new code is tested
    - [ ] Tutorials
    - [ ] Developer docs

### DIFF
--- a/src/matchbox/client/models/models.py
+++ b/src/matchbox/client/models/models.py
@@ -234,7 +234,7 @@ class Model:
         result = _handler.delete_resolution(path=self.resolution_path, certain=certain)
         return result.success
 
-    @profile()
+    @profile(attr="name")
     def run(
         self, for_validation: bool = False, cache_queries: bool = False
     ) -> ModelResults:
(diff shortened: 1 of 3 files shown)


will-langdale (approved):
This is really, really good work.

PR #397, opened 2025-12-11
Refactored handler into submodules

will-langdale:
I'm about to add a load of admin functions to the handler. This refactors to make that easier.

Handler refactored into modules.

## 👀 Guidance to review

None.

## 🤖 AI declaration

None.

## ✅ Checklist:

- [x] All new code is tested
    - [ ] Tutorials
    - [ ] Developer docs

### DIFF
--- a/src/matchbox/client/_handler/__init__.py
+++ b/src/matchbox/client/_handler/__init__.py
@@ -1,0 +1,65 @@
+"""Functions abstracting the interaction with the server API."""
+
+from matchbox.client._handler.admin import (
+    auth_status,
+    count_backend_items,
+    delete_orphans,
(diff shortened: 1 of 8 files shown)


src/matchbox/client/_handler/__init__.py
```diff
@@ -0,0 +1,65 @@
+"""Functions abstracting the interaction with the server API."""
+
+from matchbox.client._handler.admin import (
```
leo-mazzone: should we import all instead?

will-langdale: We could, but now we've been explicit, why roll back?

PR #404, opened 2025-12-16
Use streaming in client-side query

leo-mazzone:
In client-side query:
* Writing arrow batches to Parquet files
* Scanning Parquet files lazily and only collecting at the very end
* Minimising number of unnecessary operations
* Using streaming engine

Empirically, this leads in some cases to reduced memory usage, and only slightly increased processing time (about 10 seconds on our most expensive query).

## 👀 Guidance to review
**BREAKING CHANGE**: A query doesn't return source key fields anymore. This makes sense: we expect key fields to not be usable by models.

Note that the "explode" combine type is currently the most inefficient, I haven't spent any time optimising it as we never use it.

## 🤖 AI declaration
Helped generate code for writing parquet files in batches.

## ✅ Checklist:

- [x] All new code is tested
    - [ ] Tutorials
    - [ ] Developer docs

### DIFF
--- a/test/e2e/test_e2e_dag.py
+++ b/test/e2e/test_e2e_dag.py
@@ -235,11 +235,13 @@ class TestE2EPipelineBuilder:
         logging.info(f"First run produced {first_run_entities} unique entities")
 
         # Lookup works too
-        test_key = list(
-            self.linked_testkit.find_entities(
-                min_appearances={source_b.name: 1, source_a.name: 1}
-            )[0].keys[source_b.name]
-        )[0]
+        test_key = next(
+            iter(
+                self.linked_testkit.find_entities(
+                    min_appearances={source_b.name: 1, source_a.name: 1}
+                )[0].keys[source_b.name]
+            )
+        )
 
         matches = dag.lookup_key(
             from_source=source_b.name,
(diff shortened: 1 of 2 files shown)


test/e2e/test_e2e_dag.py
```diff
@@ -236,9 +235,11 @@ def test_dag_pipeline_creation_and_rerun(self) -> None:
...
 
         # Lookup works too
-        test_key = final_df.filter(
-            pl.col(source_b.f(source_b.config.key_field.name)).is_not_null()
-        )[source_b.f(source_b.config.key_field.name)][0]
+        test_key = list(
+            self.linked_testkit.find_entities(
+                min_appearances={source_b.name: 1, source_a.name: 1}
+            )[0].keys[source_b.name]
+        )[0]
```
will-langdale: Whenever I see `list()[0]` I wonder if `next()` could do the job.

```suggestion
        test_key = next(
            self.linked_testkit.find_entities(
                min_appearances={source_b.name: 1, source_a.name: 1}
            )[0].keys[source_b.name]
        )
```

leo-mazzone: Done

PR #429, opened 2026-01-08
Streamline passing exceptions from server to client

leo-mazzone:
Closes https://github.com/uktrade/matchbox/issues/297

* Created DTO for transferring error responses
* Consolidated exception middleware in API, it now solely owning the mapping of exception to error code
* Only using `ResourceOperationStatus` when no exceptions are raised, thus cutting a lot of API code
* Simplified `_handler.py` exception handling to simply reconstruct exception from DTO

## 👀 Guidance to review
**BREAKING CHANGE** Client and server need to be in sync as this changes significantly the API contract.

This is the simplest approach I could have taken (in the spirit of our development philosophy), one where exceptions are identified by a string. There are other levels with different trade-offs:

* adding validator to DTO errors to ensure the name exists in the exception registry: limited utility
* turning string to string enum generated dynamically: only additional benefit is that list of exception names are published in the API spec, which I don't think is very useful
* turning string to string enum generated statistically: we gain static validation and IDE hints, at the cost of having to manually map exceptions to string enum values, which I don't think we want to have to do

## 🤖 AI declaration

Heavy use of AI to automate repetitive parts.

## ✅ Checklist:

- [x] All new code is tested
    - [ ] Tutorials
    - [ ] Developer docs

### DIFF
--- a/src/matchbox/common/exceptions.py
+++ b/src/matchbox/common/exceptions.py
@@ -1,5 +1,6 @@
 """Custom exceptions for Matchbox."""
 
+import http
 from enum import StrEnum
 from typing import TYPE_CHECKING, Any
 
@@ -41,7 +42,7 @@ class MatchboxHttpException(MatchboxException):
     Subclasses must define http_status as a class attribute.
     """
 
-    http_status: int
+    http_status: http.HTTPStatus
 
 
 # -- Common data objects exceptions --


src/matchbox/client/_handler.py
```diff
@@ -106,52 +98,47 @@ def url_params(
...
-                # Not a custom Matchbox exception, most likely a Pydantic error
-                raise MatchboxUnparsedClientRequest(res.content)
-
-    raise MatchboxUnhandledServerResponse(
-        details=res.content, http_status=res.status_code
-    )
+    try:
+        data = res.json()
+
+        # Check if this is the new ErrorResponse format
```
will-langdale: Again, minor, but "new" is not a helpful comment in the long run. It will confuse humans and AIs alike.

src/matchbox/server/api/routers/collection.py
```diff
@@ -102,37 +89,19 @@ def create_collection(
...
-        409: {
-            "model": ResourceOperationStatus,
-            **ResourceOperationStatus.error_examples(),
-        },
-        500: {
-            "model": ResourceOperationStatus,
-            **ResourceOperationStatus.error_examples(),
-        },
+        404: {"model": ErrorResponse},
+        409: {"model": ErrorResponse},
```
will-langdale: Why did 500 go?

src/matchbox/common/exceptions.py
```diff
@@ -26,6 +27,22 @@ def __init__(self, message: str | None = None) -> None:
...
+        return None
+
+
+class MatchboxHttpException(MatchboxException):
+    """Base class for exceptions that can be transmitted over HTTP.
+
+    Subclasses must define http_status as a class attribute.
+    """
+
+    http_status: int
```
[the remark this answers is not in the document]
leo-mazzone: Done

PR #440, opened 2026-01-15
fix: configuration to override celery worker logger

sergio-recrod:
Signed-off-by: DBT pre-commit check

Override celery logger configuration so logs show what happens during processing.

- Added a `configure_celery_logging` function to:
  - Configure the root logger and set a handler for it using matchbox configuration.
  - Clear handlers from celery loggers. 
  - Propagate child logs to root so they use our configuration.
- Added a `on_celery_setup_logging` block that triggers when celery setup starts which applies the above. This only happens in celery.
- Add configuration argument to prevent celery hijacking logs.
- Unrelated to the above:
  - Added `logger.propagate = False` in `dependencies.py` to prevent duplicate log entries.
  - Corrected celery configuration argument to `worker_prefetch_multiplier`.

## 👀 Guidance to review

[This is how logs looked before](https://dbt.datadoghq.eu/logs?query=env%3Advc%20-message%3A%2A%2Fhealth%2A%20-service%3A%28matchbox-server-postgres%20OR%20matchbox-server-api%29&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=paused&storage=hot&stream_sort=time%2Cdesc&viz=stream&from_ts=1768395720000&to_ts=1768395900000&live=false)

[This is how they should look now.](https://dbt.datadoghq.eu/logs?query=env%3Advc%20-message%3A%2A%2Fhealth%2A%20-service%3A%28matchbox-server-postgres%20OR%20matchbox-server-api%29&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&messageDisplay=inline&refresh_mode=paused&storage=hot&stream_sort=time%2Cdesc&viz=stream&from_ts=1768398600000&to_ts=1768398780000&live=false)

We will be losing the process ID and the process name (main, fork...), but we will be able to see full logs of what's happening with matchbox.

Tested in dev-c running mock benchmark pipeline.

## 🤖 AI declaration

Used for initial ideas.

## ✅ Checklist:

- [x] All new code is tested
    - [ ] Tutorials
    - [ ] Developer docs

### DIFF
--- a/src/matchbox/server/uploads.py
+++ b/src/matchbox/server/uploads.py
@@ -175,8 +175,8 @@ celery.conf.update(
 
 
 @setup_logging.connect
-def on_celery_setup_logging(**kwargs):  # noqa
-    # If this signal is connected, Celery won't configure logging at all.
+def on_celery_setup_logging(**kwargs: Any) -> None:
+    """Override Celery logging configuration."""
     configure_celery_logging()
 
 


src/matchbox/server/uploads.py
```diff
@@ -164,13 +165,21 @@ def s3_to_recordbatch(
...
     # This may prevent OOM errors caused by memory not being deallocated properly
     worker_max_tasks_per_child=5,
+    # Prevent Celery from hijacking the root logger
+    worker_hijack_root_logger=False,
 )
 
 
+@setup_logging.connect
+def on_celery_setup_logging(**kwargs):  # noqa
+    # If this signal is connected, Celery won't configure logging at all.
```
will-langdale: Docstring and type hints please. `**kwargs: Any` and `-> None` should work.

will-langdale (approved):
This PR is opinionated: given we can't have both right now, we prefer matchbox logs over Celery's. Approved in the knowledge that #441 is raised to fix.

PR #468, opened 2026-01-27
Narrows types to avoid None type errors

will-langdale:
Optional types meant lots of functions could be receiving `None` but had no logic to handle that. This PR uses several mechanisms to solve this problem

Removes (draft: ~250) type checking errors.

* Adds `require()` function to narrow types, used throughout
* Uses `session.scalar().one()` to narrow types in SQLAlchemy
* Uses direct key access over `.get()` where appropriate

## 👀 Guidance to review

Review `require()` function, then go from there.

## 🤖 AI declaration

Mostly left to an AI agent to implement with human oversight.

All human reviewed.

## ✅ Checklist:

- [x] All new code is tested
    - [ ] Tutorials
    - [ ] Developer docs

### DIFF
--- a/test/server/postgresql/test_pg_sql.py
+++ b/test/server/postgresql/test_pg_sql.py
@@ -486,9 +486,7 @@ class TestGetLineage:
     ) -> None:
         """Source resolutions should return only themselves."""
         with MBDB.get_session() as session:
-            source_a = require(
-                session.get(Resolutions, 1), "Resolutions 1 not found in test fixture"
-            )  # source_a
+            source_a = session.get_one(Resolutions, 1)  # source_a
 
             lineage = source_a.get_lineage()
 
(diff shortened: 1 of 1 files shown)


src/matchbox/client/dags.py
```diff
@@ -208,13 +208,18 @@ def add_resolution(
                 validate_etl=False,
             )
         elif resolution.resolution_type == ResolutionType.MODEL:
+            # Pydantic validator ensures truth is non-None for MODEL resolutions
+            assert resolution.truth is not None
```
will-langdale: No assert in prod.

src/matchbox/client/sources.py
```diff
@@ -186,6 +186,8 @@ def sources(self) -> set[SourceResolutionName]:
     @post_run
     def to_resolution(self) -> Resolution:
         """Convert to Resolution for API calls."""
+        # @post_run decorator ensures hashes is non-None
+        assert self.hashes is not None
```
will-langdale: No assert in prod.

PR #469, opened 2026-01-28
None type narrowing for SLQAlchemy

will-langdale:
Aims to non-exaustively remove some type errors caused by optional types in the server-side code.

A slightly more formal version of #468.

Removes 99 type checking errors.

* Uses `session.get_one()` to narrow `session.get()`
* Uses `session.scalars().one()` to narrow `session.scalar()` (or some `session.execute()`)
* Narrows results that insert data with `cast()`
* Overloads `sql_to_df()` so it types correctly and uses the `QueryReturnType` `StrEnum` to tighten usage
* Adds a `require()` function to quickly wrap variables that can't be `None`

## 👀 Guidance to review

I _don't_ intend this to exhaustively remove all the errors of a certain "type" -- I aimed to knock out 100 or so errors that could be reviewed with relative ease.

The `cast(CursorResult)` is empirically right: `.rowcount` doesn't appear in the generic `Result`. Still, [docs](https://docs.sqlalchemy.org/en/21/core/connections.html#basic-usage).

Changing `MBDB.get_adbc_connection()` from `Generator[PoolProxiedConnection, Any, Any]` to `Generator[AdbcConnection, Any, Any]` shouldn't have any effect, because the context manager should still close the connection, but it makes the call sites and typing simpler.

## 🤖 AI declaration

Used to ideate. Human-written.

## ✅ Checklist:

- [x] All new code is tested
    - [ ] Tutorials
    - [ ] Developer docs

### DIFF
--- a/src/matchbox/server/postgresql/adapter/eval.py
+++ b/src/matchbox/server/postgresql/adapter/eval.py
@@ -47,6 +47,8 @@ if TYPE_CHECKING:
 
     from matchbox.server.postgresql.adapter.admin import MatchboxPostgresAdminMixin
 
+    # The evaluation mixin has self functions that reference the admin mixin
+    # This lets us refernce it in a typesafe way
     _MixinBase = MatchboxPostgresAdminMixin
 else:
     ArrowTable = Any


src/matchbox/server/postgresql/db.py
```diff
@@ -146,19 +148,26 @@ def get_session(self) -> Session:
         return self._SessionLocal()
 
     @contextmanager
-    def get_adbc_connection(self) -> Generator[PoolProxiedConnection, Any, Any]:
+    def get_adbc_connection(self) -> Generator[AdbcConnection, Any, Any]:
         """Get a new ADBC connection wrapped by a SQLAlchemy pool proxy.
 
+        The pool proxy is held within the
```
sergio-recrod: There's something missing from this docstring.

src/matchbox/server/postgresql/adapter/eval.py
```diff
@@ -43,26 +44,31 @@
 
 if TYPE_CHECKING:
     from pyarrow import Table as ArrowTable
+
+    from matchbox.server.postgresql.adapter.admin import MatchboxPostgresAdminMixin
+
+    _MixinBase = MatchboxPostgresAdminMixin
```
leo-mazzone: what's this?

src/matchbox/server/postgresql/db.py
```diff
@@ -146,19 +150,30 @@ def get_session(self) -> Session:
...
+        connection = require(
+            pool.dbapi_connection, "ADBC pool must have connection object."
+        )
 
-        conn = self._adbc_pool.connect()
         try:
-            yield conn
+            yield connection
+            connection.commit()
+        except BaseException:
```
leo-mazzone: This will not roll back if an exception is caught within the context manager. Is this intended?

PR #559, opened 2026-04-20
Parallelised unit tests and removed coverage by default

### LINKED ISSUE #407: Parallelise unit tests
Unit tests are getting slow. Parallelise them.

This likely means bringing in [`pytest-xdist`](https://pytest-xdist.readthedocs.io/en/stable/).

The bottlenecks will be:

* Using multiple databases within the backend and warehouse concurrently
* The deployed API in Docker

## Databases

AI reckons something like this will do the trick for the database side of things:

```python
# In test/fixtures/db.py

@pytest.fixture(scope="session")
def worker_id(request):
    """Get the worker id (e.g., gw0, gw1) or empty string if not using xdist."""
    if hasattr(request.config, "workerinput"):
        return request.config.workerinput["workerid"]
    return ""

@pytest.fixture(scope="session")
def matchbox_postgres_settings(
    development_settings: DevelopmentSettings,
    matchbox_datastore: MatchboxDatastoreSettings,
    worker_id: str,  # Inject worker_id
) -> MatchboxPostgresSettings:
    # Append worker_id to database name to isolate workers
    db_name = "matchbox" if not worker_id else f"matchbox_{worker_id}"
    
    return MatchboxPostgresSettings(
        batch_size=250_000,
        postgres={
            # ... keep other settings ...
            "database": db_name, 
            # ...
        },
        datastore=matchbox_datastore,
    )

# Do the same for postgres_warehouse fixture to isolate the source data
@pytest.fixture(scope="function")
def postgres_warehouse(
    development_settings: DevelopmentSettings,
    worker_id: str,
) -> Generator[Engine, None, None]:
    # ...
    database = "warehouse" if not worker_id else f"warehouse_{worker_id}"
    # ...
```

## API

Tricker.

If [FastAPI's `TestClient`](https://fastapi.tiangolo.com/tutorial/testing/) is just an in-process version of the actual API, you might consider it appropriate to use  it in the e2e tests.

However, you might still want to run e2e tests with the deployed API container in the PR tests.

You might also consider that running e2e tests in parallel is actually a _feature_, not a bug, and the lack of isolation will create stronger tests conditions.

Not sure on direction with this one.

## Acceptance

Unit tests run faster due to parallelisation.

will-langdale:
Parallelises unit tests where relatively easy, marking some tests as serial only.

Reduces test time on my personal machine from 4m to 1m 30s. Removes coverage for an extra 10s speedup (1m 19s) as it's never used. Still accessible with `just test default --cov`.

Closes #407

### Benchmarks

* Ryzen 7 7840HS, Fedora: 1m 19s
* MacBook Pro M3: 1m 21s
* GitHub runner, 4-CPU Ubuntu, no idea of clock: 1m 35s

A ~15s difference between a runner and an M3 ain't bad at all!

* Adds `MatchboxDatabase.settings_scope()` -- 15 lines that makes worker-specific instances easy to run
* Modifies fixtures to allow workers
* Adds pytest-xdist as a dependency

Also

* Adds [false positive exclusions](https://github.com/uktrade/github-standards?tab=readme-ov-file#excluding-false-positives-1) for Presidio as we were failing GitHub Security Standards
* Updates the uv lockfile to placate dependabot and MkDocs (some issue with a Griffe import)

## 👀 Guidance to review

None.

## 🤖 AI declaration

AI used throughout, all human reviewed.

## ✅ Checklist:

- [x] All new code is tested
    - [ ] Tutorials
    - [ ] Developer docs


test/e2e/test_e2e_evaluation.py
```diff
@@ -23,6 +23,8 @@
 
 
 @pytest.mark.docker
+@pytest.mark.serial
+@pytest.mark.xdist_group("serial")
```
leo-mazzone, line 27: what's the point of the non-xdist marker?

test/fixtures/db.py
```diff
@@ -292,11 +388,14 @@ def upload_tracker_in_memory() -> Generator[InMemoryUploadTracker, None, None]:
 @pytest.fixture(scope="function")
 def upload_tracker_redis(
     development_settings: DevelopmentSettings,
+    worker_id: str,
```
leo-mazzone, line 410: is this automatically injected by xdist?

test/e2e/test_e2e_evaluation.py
will-langdale: `.xdist_group()` doens't go into `pytest -m`. The extra mark exists so you can run slower serial tests separately.

test/fixtures/db.py
will-langdale: Yes, it's an [xdist fixture](https://pytest-xdist.readthedocs.io/en/stable/how-to.html#identifying-the-worker-process-during-a-test).

test/fixtures/db.py
```diff
@@ -204,59 +285,74 @@ def matchbox_datastore(
...
 
 
-@pytest.fixture(scope="function")
-def matchbox_postgres(
-    matchbox_postgres_settings: MatchboxPostgresSettings,
+@pytest.fixture(scope="session")
+def matchbox_postgres_session(
+    development_settings: DevelopmentSettings,
+    matchbox_datastore: MatchboxDatastoreSettings,
+    worker_id: str,
```
leo-mazzone, line 316: There are now 4 fixtures:
- postgres_warehouse_database
- matchbox_postgres_session
- matchbox_postgres
- shared_matchbox_postgres

It's hard to understand the difference. I'd like you to rewrite all docstrings to make it obvious what each is for

will-langdale: Done.

leo-mazzone (approved):

PR #590, opened 2026-06-08
Parameterise PostgreSQL schema throughout the backend

will-langdale:
`MB__SERVER__POSTGRES__DB_SCHEMA` was only respected by the SQLAlchemy ORM metadata. All Alembic migrations, several raw SQL strings, and the ingest utilities were hardcoded to "mb".

## 🛠️ Changes proposed in this pull request

- `env.py`: pass `db_schema` into Alembic config so migrations can read it
- All 25 migration files: replace `schema="mb"` kwargs, inline FK strings (e.g. `["mb.table.col"]`), and raw SQL table references with the parameterised value via `op.get_context().config.get_main_option()`
- Adds an Alembic migration hook to ensure this happens in future migrations
- First migration: remove `CREATE`/`DROP SCHEMA` — the operator is now responsible for creating the schema before running migrations
- `db.py`: parameterise the pre-alembic recovery DROP SCHEMA
- `orm.py`: fix hardcoded schema on the upload_stages Enum type
- `insert.py`, adapter/eval.py: fix `schema_name="mb"` on temp table calls
- `justfile`: use `$MB__SERVER__POSTGRES__DB_SCHEMA` in the drop recipe

## 👀 Guidance to review

Alembic post migration generation hook in action:

```
(matchbox-db) willlangdale@DBT001350 ~/D/matchbox (claude/awesome-mccarthy-p5Eb8)> just migrate generate "Add level_2."
uv run alembic --config "src/matchbox/server/postgresql/alembic.ini" revision --autogenerate -m "Add level_2."
...
INFO  [alembic.autogenerate.compare.tables] Detected added column 'mb.step_from.level_2'
  Generating /Users/willlangdale/Dev/matchbox/src/matchbox/server/postgresql/alembic/versions/5e33e5844258_add_level_2.py ...  done
  Running post write hook 'fix_schema' ...
  done
  Running post write hook 'ruff_format' ...
1 file reformatted
  done
  Running post write hook 'ruff_lint' ...
Found 2 errors (2 fixed, 0 remaining).
  done
```

```python
"""Add level_2.

Revision ID: 5e33e5844258
Revises: 65dd8d1fe119
Create Date: 2026-06-09 13:11:20.645400

"""

from collections.abc import Sequence

import sqlalchemy as sa
from alembic import op

# revision identifiers, used by Alembic.
revision: str = "5e33e5844258"
down_revision: str | None = "65dd8d1fe119"
branch_labels: str | Sequence[str] | None = None
[...]

PR #605, opened 2026-07-02
Alembic always migrates

will-langdale:
When we removed schema management in #590 we left in a place to create the schema which flouts the "db manager creates the schema" rule: when migrations run for the first time.

* Removes that line then fixes tests that need the schema.
* Adds some docstrings cause I like 'em.
* Deals with a warning that polars is about to change the default arg for `explode(empty_as_null)` by making the default value explicit

## 👀 Guidance to review

None.

## 🤖 AI declaration

Ideated with AI. Implemented by hand.

## ✅ Checklist:

- [x] All new code is tested
    - [ ] Tutorials
    - [ ] Developer docs


a-gleeson (approved):
LGTM!

Issue #560, opened 2026-04-27
Make matchbox work local-only


will-langdale:
From observing and onboarding users, we know that many analytical users want to work local-only, and don't need the API.

We will eventually want online matching, where we match one new record at a time. For this to be quick, we'll need to avoid making API calls we don't need to.

Both use cases mean making matchbox work local-only.

This issue is a place to work through the implementation, and track process.

## Acceptance

Matchbox DAGs can run end to end without talking to the API.

will-langdale:
This is a plan I worked up in conjunction with AI.

## Aim

Run matchbox pipelines entirely client-side — Source indexing, Model scoring, and Resolver clustering — with no server dependency during execution. Sync results to the server as a separate, optional step.

This means a user can:

1. Define a DAG
2. Call `dag.run()` — everything executes locally, all intermediate state on disk
3. Inspect results
4. Call `dag.sync()` — uploads results to the server in the background
5. Or not sync at all

Additionally, `dag.load_default()` and `dag.load_pending()` pull a server-side run into local state, enabling local querying and inspection of results that someone else produced.

## The Core Problem

The current architecture treats the server as the source of truth for identity. Every cluster in the system gets a PostgreSQL-assigned `BIGINT` autoincrement ID. These IDs flow through every stage:

- `Source.sync()` uploads hashes → server assigns `cluster_id`s
- `Query.data_raw()` calls the server → gets `(cluster_id, key)` mappings back
- Models produce edges referencing server `cluster_id`s
- Resolvers produce clusters referencing server `cluster_id`s

The critical bottleneck is `Query.data_raw()`. It makes an HTTP call to the server to resolve "which integer ID does this source row have?" — and models cannot run without this mapping. Every other server dependency follows from this one.

## The Solution: Content-Addressed Identity with Local Integer Interning

Every cluster is already uniquely identified by a content hash — the server stores `cluster_hash` alongside `cluster_id`, and `hash_cluster_leaves()` in `common/transform.py` computes deterministic hashes for resolver clusters from their members. The hash is the true identity; the integer ID is an optimisation for joins and ER algorithms.

The solution is an **interner**: a structure that assigns dense local `uint64` IDs to content hashes, providing cheap bidirectional lookup. ER algorithms work with integers in their hot path (as they must). The interner is the boundary between content-addressed identity and integer-optimised computation.

[...]

PR #614, opened 2026-07-16
Factor common SQL out of backend adapter to enable eventual local adapter

will-langdale:
Part of #560. Covers PRs A-C.

A local adapter and a server adapter will have some common machinery, and some distinct. This factors out the common stuff, and refactors unit tests so they can cover both types of adapter.

Moves common SQLAlchemy tables and functions into `common`, then uses them in the PostgreSQL adapter.

With the tables, we use the [Declarative with Imperative Table (Hybrid Declarative) pattern](https://docs.sqlalchemy.org/en/20/orm/declarative_tables.html#declarative-with-imperative-table-a-k-a-hybrid-declarative) so we have shared table definitions that can be used in different databases.

A side effect of this is we have one shared symbolic `"mb"` metadata plus `schema_translate_map` at the engine. I've commented it when it's present as it's confusing to see at a low level.

I appreciate the diff is hard when it's 90% lift and shift, so these are the ways in which the moved functions differ:

* Lineage computed by the caller now (`step.get_lineage()`), builders take it as a plain argument instead of calling it themselves internally.
* `match()` computes lineage once and reuses it for both CTEs; originally each CTE independently recomputed identical lineage via two DB round trips.
* `_compute_resolver_hashes` split into three pieces: `build_expanded_leaves_subquery`, a new `build_leaf_hash_groups_query`, and pure-Python `hash_resolver_parents`.
* `hash_resolver_parents` takes `Iterable[Row]` not `Iterable[tuple[int, Iterable[bytes]]]`
* `clear_database()` schema lookup fixed: `self.MatchboxBase.metadata.schema` (now symbolic) swapped for `self.settings.postgres.db_schema`

## 👀 Guidance to review

Are you happy with the high level SQLAlchemy pattern I've used? Alternatives were to make factories, but I like this better. I've also gone with classlike var names which technically violates PEP8 but I think reads way nicer.

## 🤖 AI declaration

AI written, human reviewed.

## ✅ Checklist:

- [x] All new code is tested
    - [ ] Tutorials
    - [ ] Developer docs

### DIFF
--- a/src/matchbox/common/adapters/__init__.py
+++ b/src/matchbox/common/adapters/__init__.py
@@ -1,0 +1,7 @@
+"""Shared backend adapter contracts and machinery.
+
+Code shared by every Matchbox backend: the cluster store protocol ABC,
+and the SQL machinery that relational backends compose to implement it.
+
+TODO: local client-side backends will depend on this package too.
+"""
--- a/src/matchbox/common/adapters/__init__.py
+++ b/src/matchbox/common/adapters/__init__.py
@@ -2,6 +2,4 @@
 
 Code shared by every Matchbox backend: the cluster store protocol ABC,
 and the SQL machinery that relational backends compose to implement it.
-
-TODO: local client-side backends will depend on this package too.
 """
--- a/src/matchbox/common/adapters/protocol.py
+++ b/src/matchbox/common/adapters/protocol.py
@@ -1,9 +1,12 @@
 """The cluster store protocol shared by every Matchbox backend."""
 
+import json
 from abc import ABC, abstractmethod
-from typing import Protocol
+from enum import StrEnum
+from typing import Any, Protocol
 
 from pyarrow import Table
+from pydantic import BaseModel, field_validator
 
 from matchbox.common.dtos import (
     Match,
@@ -37,6 +40,29 @@ class ListableAndCountable(Countable, Listable):
     pass
 
 
+class MatchboxBackends(StrEnum):
+    """The available backends for Matchbox."""
+
+    POSTGRES = "postgres"
+
+
+class MatchboxSnapshot(BaseModel):
+    """A snapshot of the Matchbox database."""
+
+    backend_type: MatchboxBackends
+    data: dict[str, Any]
+
+    @field_validator("data")
+    @classmethod
+    def check_serialisable(cls, value: dict[str, Any]) -> dict[str, Any]:
+        """Validate that the value can be serialised to JSON."""
+        try:
+            json.dumps(value)
+            return value
+        except (TypeError, OverflowError) as e:
+            raise ValueError(f"Value is not JSON serialisable: {e}") from e
+
+
 class MatchboxClusterStoreAdapter(ABC):
     """The contract shared by anything that stores and serves cluster data.
 
@@ -50,9 +76,9 @@ class MatchboxClusterStoreAdapter(ABC):
     source_clusters: Countable
     model_clusters: Countable
     all_clusters: Countable
-    creates: Countable
-    merges: Countable
-    proposes: Countable
+    creates: Countable  # clusters proposed by resolver
+    merges: Countable  # cluster lineage tree
+    proposes: Countable  # raw model edge scores
 
     # Query block
 
@@ -143,4 +169,46 @@ class MatchboxClusterStoreAdapter(ABC):
     @abstractmethod
     def get_resolver_data(self, path: ResolverStepPath) -> Table:
         """Get cluster assignments for a resolver step."""
+        ...
+
+    # Data management
+
+    @abstractmethod
+    def dump(self) -> MatchboxSnapshot:
+        """Dumps the entire database to a snapshot.
+
+        Returns:
+            A MatchboxSnapshot with the database's current state, tagged
+                with the backend type that produced it.
+        """
+        ...
+
+    @abstractmethod
+    def restore(self, snapshot: MatchboxSnapshot) -> None:
+        """Restores the database from a snapshot.
+
+        Args:
+            snapshot: A MatchboxSnapshot with the database's state
+
+        Raises:
+            TypeError: If the snapshot's backend type doesn't match this one
+        """
+        ...
+
+    @abstractmethod
+    def clear(self, certain: bool) -> None:
+        """Soft clear the database by deleting all rows but retaining tables.
+
+        Args:
+            certain: Whether to delete the database without confirmation.
+        """
+        ...
+
+    @abstractmethod
+    def drop(self, certain: bool) -> None:
+        """Hard clear the database by dropping all tables and re-creating.
+
+        Args:
+            certain: Whether to drop the database without confirmation.
+        """
         ...
--- a/src/matchbox/common/adapters/protocol.py
+++ b/src/matchbox/common/adapters/protocol.py
@@ -1,0 +1,146 @@
+"""The cluster store protocol shared by every Matchbox backend."""
+
+from abc import ABC, abstractmethod
+from typing import Protocol
+
+from pyarrow import Table
+
+from matchbox.common.dtos import (
+    Match,
+    ModelStepPath,
+    ResolverStepPath,
+    SourceStepPath,
+    Step,
+    StepPath,
+)
+
+
+class Countable(Protocol):
+    """A protocol for objects that can be counted."""
+
+    def count(self) -> int:
+        """Counts the number of items in the object."""
+        ...
+
+
+class Listable(Protocol):
+    """A protocol for objects that can be listed."""
+
+    def list_all(self) -> list[str]:
+        """Lists the items in the object."""
+        ...
+
+
+class ListableAndCountable(Countable, Listable):
+    """A protocol for objects that can be counted and listed."""
+
+    pass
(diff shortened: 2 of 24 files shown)


src/matchbox/common/adapters/sql/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""Shared SQL machinery for relational Matchbox backends.
+
+Dialect-neutral Core table metadata, query and match builders, and
+resolver canonicalisation logic for the five cluster tables.
```
leo-mazzone: don't say there's five

src/matchbox/common/adapters/protocol.py
```diff
@@ -0,0 +1,146 @@
...
+    model_clusters: Countable
+    all_clusters: Countable
+    creates: Countable
+    merges: Countable
+    proposes: Countable
+
+    # Query block
+
+    @abstractmethod
+    def query(
```
leo-mazzone, line 86: do we want for the server to retain the ability to query? In the potential alternative discussed yesterday, the local engine wouldn't need to query and I thought we could get rid of this whole method everywhere as a result. But I guess we're saying we want to retain the ability to query the server? Why though? I understand why you might want to be able to match() on the server, because you could do a live ID lookup without having to download a bunch of stuff locally. There is a product question here to ponder

src/matchbox/common/adapters/protocol.py
```diff
@@ -0,0 +1,146 @@
...
+    implement this too.
+    """
+
+    # Counts
+    source_clusters: Countable
+    model_clusters: Countable
+    all_clusters: Countable
+    creates: Countable
+    merges: Countable
+    proposes: Countable
```
leo-mazzone: is this an opportunity for making these less confusing?

src/matchbox/common/adapters/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""Shared backend adapter contracts and machinery.
+
+Code shared by every Matchbox backend: the cluster store protocol ABC,
+and the SQL machinery that relational backends compose to implement it.
+
+TODO: local client-side backends will depend on this package too.
```
leo-mazzone: what's the todo here?

src/matchbox/common/adapters/sql/tables.py
```diff
@@ -0,0 +1,157 @@
...
+Follows the declarative with imperative table pattern: these are plain
+SQLAlchemy Core Table objects on one shared MetaData, used directly by
+every backend. Relational backends adopt them via __table__ on their own
+ORM classes, or use them as-is with Core.
+
+Rules for this module:
+
+- Generic types only (BigInteger, LargeBinary, Text, REAL). No
+    dialect-specific types, no native enums. If a column genuinely needs a
+    dialect type, use type.with_variant(PGType, "postgresql") on the shared
```
leo-mazzone, line 12: if you did that, how could that column still be in common?

src/matchbox/common/adapters/sql/tables.py
```diff
@@ -0,0 +1,157 @@
...
+    ForeignKey,
+    Index,
+    LargeBinary,
+    MetaData,
+    Table,
+    Text,
+    UniqueConstraint,
+)
+
+metadata = MetaData(schema="mb")
```
leo-mazzone: perhaps ALL_CAPS? Just for style, since it's being imported by other modules

src/matchbox/server/postgresql/orm.py
```diff
@@ -21,6 +20,7 @@
 from sqlalchemy.dialects.postgresql import BYTEA, JSONB, TEXT, insert
 from sqlalchemy.orm import Mapped, Session, mapped_column, relationship, selectinload
```
leo-mazzone, line 21: is there any performance or compatibility gain in using Postgres-native types? Asking to understand the impact of the common tables

src/matchbox/common/factories/scenarios.py
```diff
@@ -199,26 +199,34 @@ def create_preindex_scenario(
     The warehouse contains three interlinked tables that cover common linkage
     realistic scenarios, but are not yet indexed.
     """
-    # First create the admin scenario
-    dag_testkit = create_admin_scenario(
```
leo-mazzone, line 203: what's happened here?

will-langdale: All the things added in the admin scenario are specific to the server, so instead we import the bare scenario then logic gate the admin additions.

src/matchbox/server/postgresql/orm.py
will-langdale: All the types in common translate to what we want them to in PostgreSQL dialect, so concretely no right now. The `type.with_variant(PGType, "postgresql")` can give us the type we need if it every wasn't the case.

src/matchbox/common/adapters/sql/tables.py
will-langdale: Why not? It's specifically there so the type would be whatever it needs to be when it's attached to a specific ORM system. It's like saying "you're `sa.BigInteger`, but make sure you translate that to psql.Double if attached to PostgreSQL" (as an inaccurate but illustrative example).

src/matchbox/common/adapters/__init__.py
will-langdale: Was trying to show where future stuff would go. I'll remove.

src/matchbox/common/adapters/protocol.py
will-langdale: Feels out of scope for a lift and shift, unless you have some specific, obvious change we both agree on. Maybe this is a broader question? Can you raise an issue if so?

src/matchbox/common/adapters/protocol.py
will-langdale: My rough thinking was:

* Query's sort of like a batch match (and will be moreso when you don't need to do it once per source)
* It's quite expensive to download a whole DAG to perform a query/match operation -- can imagine thin scripts using the endpoints
* Free to keep as-is

These feel weaker written down tbh.

What do you think?

will-langdale: I thought about this over the weekend. I can definitely see us removing query from the API, or even changing its role. I also think there's uncertainty about the local performance of query that we'd want to bear out: what's the impact on speed, memory, how does the UX feel, etc.

Bearing in mind this uncertainty I'd like to add this to the issue comments along with the existing performance one, and revisit once the feature is more complete.

leo-mazzone: when you do this, I'll resolve

src/matchbox/common/adapters/protocol.py
leo-mazzone: At a minimum, I'd like to read this and understand at a glance the difference. Be it my sleep deprivation, but what is each one? Creates are new clusters, proposes are probability rows and merges are (?)

src/matchbox/common/adapters/sql/tables.py
will-langdale: For reference: [`with_variant()` documentation](https://docs.sqlalchemy.org/en/21/core/type_api.html#sqlalchemy.types.TypeEngine.with_variant).

src/matchbox/common/adapters/protocol.py
will-langdale: [Added.](https://github.com/uktrade/matchbox/issues/560#issuecomment-5021664742)

src/matchbox/common/adapters/protocol.py
will-langdale: Added comments.

src/matchbox/common/factories/scenarios.py
will-langdale: I've reworked the whole scenario file so the situation is clearer.

Scenarios must now always include their _state_ and whether they apply to **server** backends only or **either**. They with error if the wrong one is used.

I think there's likely little issues with what I've done that will be hard to work through until the next PR, and we have an actual local backend to test.

src/matchbox/common/adapters/protocol.py
```diff
@@ -37,6 +40,29 @@ class ListableAndCountable(Countable, Listable):
...
+
+    backend_type: MatchboxBackends
+    data: dict[str, Any]
+
+    @field_validator("data")
+    @classmethod
+    def check_serialisable(cls, value: dict[str, Any]) -> dict[str, Any]:
+        """Validate that the value can be serialised to JSON."""
+        try:
+            json.dumps(value)
```
leo-mazzone, line 60: how often would this be done?

src/matchbox/common/factories/scenarios.py
will-langdale: Discussed: slight wording change pushed for clarity.

src/matchbox/common/adapters/protocol.py
will-langdale: Discussed: we agree the wording here is increasingly murky and drifting from its original intent. Raised as an issue in #622 to bring it back to something robust and useful.

leo-mazzone (approved):

PR #615, opened 2026-07-16
fix: make resolver priority deterministic

### LINKED ISSUE #612: Resolver priority depends on database insertion order
issue: 
- resolver query priority is currently ordered by minimum lineage distance then step_id
- distance is part of the logical graph
- step_id is an auto-incrementing database identifier determined by insertion order

so two equivalent dags can produce different query results:
- two resolver ancestors are at the same minimum distance
- both claim the same leaf
- they assign it to different roots
- no nearer resolver claims it

this is an issue for https://github.com/uktrade/matchbox/issues/560 because of the following:
- local execution can't reproduce server behaviour before syncing because no server step_id exists
- users can't inspect or intentionally control the tie-break

i.e., given the same dag and resolver outputs,  we need query resolution to be independent of database insertion order

options:
1. define an explicit resolver priority
2. (recommended) define a deterministic dag-derived tie-break (since names are restricted to ascii characters, python lexical ordering and postgres c collation over step names should agree)
3. reject competing equal-depth opinions as ambiguous

joshwong-cs:

Resolver priority for steps at the same lineage depth was ordered by `step_id`. Since step IDs are assigned during insertion, equivalent runs could produce different resolver priorities.

Order equal-depth steps by name using PostgreSQL's `C` collation instead. Since step names are restricted to ASCII characters, it should agree with Python's lexical ordering and give the "server-less" portability that https://github.com/uktrade/matchbox/issues/560 requires.

Lineage depth remains the primary ordering criterion.

Closes [#612](https://github.com/uktrade/matchbox/issues/612).

## 🤖 AI declaration

None

## ✅ Checklist:

- [x] All new code is tested
    - [ ] Tutorials
    - [ ] Developer docs

### DIFF
--- a/src/matchbox/server/postgresql/orm.py
+++ b/src/matchbox/server/postgresql/orm.py
@@ -330,7 +330,7 @@ class Steps(CountMixin, MBDB.MatchboxBase):
     ) -> list[tuple[int, int | None]]:
         """Returns lineage ordered by priority.
 
-        Highest priority (lowest level) first, then by step_id for stability.
+        Highest priority (lowest level) first, then by step name for stability.
 
         Args:
             sources: If provided, only return lineage paths that lead to these sources
@@ -386,7 +386,7 @@ class Steps(CountMixin, MBDB.MatchboxBase):
                 )
 
             results = session.execute(
-                query.order_by(StepFrom.level.asc(), StepFrom.parent.asc())
+                query.order_by(StepFrom.level.asc(), Steps.name.collate("C").asc())
             ).all()
 
             # Get self's source config ID


test/server/postgresql/test_pg_core.py
```diff
@@ -57,3 +59,53 @@ def test_ingest_to_temporary_table(
         pytest.raises(Exception),  # Should fail as table is dropped # noqa: B017
     ):
         session.execute(select(func.count()).select_from(temp_table))
+
+
+@pytest.mark.docker
+def test_lineage_tie_breaks_by_step_name(
```
will-langdale: This is an objection as much about culture as it is about good engineering practise, and is very much as discussion.

I'm really conscious of overtesting. I think that this change is small, clear, defers to a well-tested subsystem and if it's covered, it should be either as a trivial parameter to another test, or as an additional assertion to something that already exists. I personally wouldn't have added a test.

Each new test costs compute time and human attention. I don't think this change is big enough or the regression impactful enough to justify paying the cost of this specific test.

You wrote the test, so you clearly feel it needs coverage. Is there a way you can cover it that requires 2 lines rather than 50?

will-langdale (changes requested):

test/server/postgresql/test_pg_core.py
joshwong-cs: had a look around:
- adapter query tests pull lineage indirectly
- none create conflicting equal-depth resolver assignments where ordering changes the result
- making one do that would probably require substantial fixture changes (we might lose clarity on existing tests)

happy to remove test, but maybe we should look into augmenting our subsystem tests to ensure/prove deterministic resolution of things like these:
- multiple equal-depth resolver ancestors
- overlapping resolver assignments
- conflicting cluster assignments
- equivalent dags constructed in different orders
- lineage tie-breaker

will-langdale: If you raise that as an issue I'll happily approve this PR.

joshwong-cs: https://github.com/uktrade/matchbox/issues/619

will-langdale (approved):

PR #617, opened 2026-07-17
fix: prevent idle transactions in the ADBC pool

### LINKED ISSUE #616: Connection pooling keeps idle connections in transactions
In the PostgreSQL adapter, connection pooling keeps idle connections in transactions, so long-lived pipelines are causing errors.

I suspect this has only become apparent as we've migrated to a new PostgreSQL instance where we don't have such fine-grained control of DB settings. Worth splatting the bug, if you can reproduce.

## Acceptance

Fix it.

joshwong-cs:

ADBC disables auto-commit by default and immediately starts another transaction after commit or rollback. Returning those connections to the pool leaves every connection idle in transaction. PostgreSQL eventually terminates those connections when `idle_in_transaction_session_timeout` is reached.

Create each physical pool connection directly with auto-commit enabled instead of cloning a transactional source connection. Invalidate pool entries that raise ADBC errors so terminated connections are not reused. Matchbox’s normal SQLAlchemy sessions still use explicit transactions, i.e., the change only affects the ADBC pool and does not change the public API.

Fixes https://github.com/uktrade/matchbox/issues/616.

## 👀 Guidance to review

Running this on main against db shows the API's supposedly unused ADBC connections still have active transactions:

```sql
# you should see a bunch of these: state | idle in transaction
SELECT pid, state, xact_start, state_change, query FROM pg_stat_activity
WHERE datname = current_database() ORDER BY pid;
```

## 🤖 AI declaration

None.

## ✅ Checklist:

- [x] All new code is tested
    - [ ] Tutorials
    - [ ] Developer docs

### DIFF
--- a/src/matchbox/server/postgresql/db.py
+++ b/src/matchbox/server/postgresql/db.py
@@ -4,7 +4,7 @@ import threading
 from collections.abc import Generator
 from contextlib import contextmanager
 from pathlib import Path
-from typing import Any
+from typing import Any, cast
 
 from adbc_driver_postgresql import dbapi as adbc_dbapi
 from adbc_driver_postgresql.dbapi import Connection as AdbcConnection
@@ -19,6 +19,7 @@ from sqlalchemy import (
     create_engine,
     text,
 )
+from sqlalchemy.engine.interfaces import DBAPIConnection
 from sqlalchemy.orm import Session, declarative_base, sessionmaker
 from sqlalchemy.pool import QueuePool
 
@@ -80,7 +81,6 @@ class MatchboxDatabase:
         self._SessionLocal: sessionmaker | None = None
         self._adbc_pool: QueuePool | None = None
         self._adbc_lock = threading.Lock()
-        self._source_adbc_connection: adbc_dbapi.Connection | None = None
         self.MatchboxBase = declarative_base(
             metadata=MetaData(schema=settings.postgres.db_schema)
         )
@@ -113,20 +113,20 @@ class MatchboxDatabase:
             self._SessionLocal = None
 
     def _connect_adbc(self) -> None:
-        self._source_adbc_connection = adbc_dbapi.connect(
-            self.connection_string(driver=False)
-        )
+        # adbc clones default to non-autocommit and stay idle in transaction
         self._adbc_pool = QueuePool(
-            self._source_adbc_connection.adbc_clone,
+            lambda: cast(
+                DBAPIConnection,
+                adbc_dbapi.connect(
+                    self.connection_string(driver=False), autocommit=True
+                ),
+            ),
         )
 
     def _disconnect_adbc(self) -> None:
         if self._adbc_pool:
             self._adbc_pool.dispose()
             self._adbc_pool = None
-
-            self._source_adbc_connection.close()
-            self._source_adbc_connection = None
 
     @contextmanager
     def settings_scope(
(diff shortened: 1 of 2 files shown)


test/server/postgresql/test_pg_core.py
```diff
@@ -1,13 +1,35 @@
...
+from sqlalchemy import func, select, text
 from sqlalchemy.dialects.postgresql import BIGINT, TEXT
 
 from matchbox.server.postgresql import MatchboxPostgres
 from matchbox.server.postgresql.db import MBDB
 from matchbox.server.postgresql.utils.db import ingest_to_temporary_table
 
 
+@pytest.mark.docker
+def test_adbc_pool_connection_is_not_idle_in_transaction(
```
will-langdale: Culture challenge: we prefer short, with detail in the docstring. How we phrase it in CDL's `AGENTS.md`:

>In tests, favour short explanatory names that only signal the test's role in the suite. Put the extra explanation in the docstring.

src/matchbox/server/postgresql/db.py
```diff
@@ -113,21 +113,21 @@ def _disconnect(self) -> None:
             self._SessionLocal = None
 
     def _connect_adbc(self) -> None:
-        self._source_adbc_connection = adbc_dbapi.connect(
-            self.connection_string(driver=False)
-        )
+        # adbc clones default to non-autocommit and stay idle in transaction
         self._adbc_pool = QueuePool(
-            self._source_adbc_connection.adbc_clone,
```
will-langdale, line 120: Cloning is the way [their docs advise doing connection pooling](https://arrow.apache.org/adbc/current/python/recipe/postgresql.html#connection-pooling-with-sqlalchemy).

I don't fully know why that's the case, but I suspect it's faster when cycling lots of connections. Could something like this give us the base of both worlds:

```python
def _connect_adbc(self) -> None:
    self._source_adbc_connection = adbc_dbapi.connect(
        self.connection_string(driver=False), autocommit=True
    )
    def _make_clone():
        clone = self._source_adbc_connection.adbc_clone()
        clone.adbc_connection.set_autocommit(True)
        return clone
    self._adbc_pool = QueuePool(_make_clone)
```

Also worth noting is they're [quite receptive to changes upstream](https://redirect.github.com/apache/arrow-adbc/issues/2673). Do you think it's worth us raising an issue so that either clone _does_ copy the autocommit, or it's mentioned in the recipe that clones reset it?

will-langdale (changes requested):

src/matchbox/server/postgresql/db.py
joshwong-cs: tried it and `QueuePool` calls rollback() by default when a connection is returned to the pool:
1. the low-level connection is in auto-commit mode
2. the DB-API wrapper still thinks rollback is supported
3. QueuePool calls clone.rollback()
4. ADBC forwards it to the low-level connection
5. PostgreSQL ADBC rejects it

i don't think its worth the low-level hassle given that connections are probably not being constantly created and destroyed? e.g. process startup, pool expansion, database failover, genuine network errors, connection invalidation?

but yeah i agree best long-term solution belongs upstream in ADBC

test/server/postgresql/test_pg_core.py
joshwong-cs: :robot:

src/matchbox/server/postgresql/db.py
will-langdale: I agree with the point on hassle. 

I think it would be appreciated and neighbourly if you raised the issue upstream, but I'm not going to ask you to as part of this PR.

Feel free to resolve this conversation once you've read it.

will-langdale (approved):

PR #625, opened 2026-07-21
Adds duckdb-backed local adapter

will-langdale:
Adds a duckdb-backed local adapter.

Part of #560.

* Adds a local adapter ABC
* Adds duckdb-backed adapter class
* Adds duckdb ORM and tests. Note that autoincrement keys needed some interesting piping

## 👀 Guidance to review

I tried to factor out more stuff into the common SQL module. Do you agree with my decisions? Did I go too far, or not far enough? The principle I was trying to follow was to keep common subqueries separate to where they're hooked up to specific ORM tables or dialect mechanics. This is why inner functions of `insert` are abstracted rather than the whole thing.

Are you happy with how components in the two ORMs are shared? I was sceptical of the ORM as there's some repeated code in things like `SourceConfigs`, but given the alternatives (a different type of mixins, subclasses), this way seems smallest, seeing your ORM in one place is good, and I didn't want to add leaky abstractions.

Do you agree with the decision not to have a package duckdb singleton, so you can create lots of instances if you wish? We can't control what the user does and how they import like we do on the server.

Is it time to review how resolvers are stored locally, simplify query's logic, and cut it entirely from the server? Or is it simpler to mirror the logic here and pick it up later? Two facets: is it the right move AND will is swell the PR?

We implement `_cascade_invalidate` because duckdb (rightly) lacks `delete cascade`: it's just not made for that type of OLTP type work.

## 🤖 AI declaration

AI written but human reviewed throughout. Went through 2-3 iterations and spikes to get to this shape, including toys to validate duckdb mechanisms like sequences.

## ✅ Checklist:

- [x] All new code is tested
    - [ ] Tutorials
    - [ ] Developer docs

### DIFF
--- a/src/matchbox/client/adapters/duckdb/adapter.py
+++ b/src/matchbox/client/adapters/duckdb/adapter.py
@@ -314,7 +314,7 @@ class MatchboxLocalDuckDBQueryMixin:
 
 
 class MatchboxLocalDuckDBDataMixin:
-    """Data mixin for the local DuckDB adapter: protocol block, replace policy."""
+    """Data mixin for the local DuckDB adapter."""
 
     def create_step(self, step: Step, path: StepPath) -> None:  # noqa: D102
         with self._session() as session:
@@ -567,12 +567,12 @@ class MatchboxLocalDuckDBDataMixin:
         db.drop_dynamic_cache_tables(
             self._engine, (_RAW_DATA_TABLE_PREFIX, _QUERY_CACHE_TABLE_PREFIX)
         )
-        db.drop_all_duckdb(self._engine, [*orm.SHARED_TABLES, *orm.LOCAL_TABLES])
-        db.create_all_duckdb_safe(self._engine, orm.LOCAL_TABLES)
-        db.create_all_duckdb_safe(self._engine, orm.SHARED_TABLES)
+        db.drop_db(self._engine, [*orm.SHARED_TABLES, *orm.LOCAL_TABLES])
+        db.create_db(self._engine, orm.LOCAL_TABLES)
+        db.create_db(self._engine, orm.SHARED_TABLES)
 
 
-class MatchboxLocalDuckDBLocalMixin:
+class MatchboxLocalDuckDBCacheMixin:
     """Local-only mixin: raw data, query cache, cascade invalidation."""
 
     def insert_raw_data(self, path: SourceStepPath, table: ArrowTable) -> None:  # noqa: D102
@@ -712,7 +712,7 @@ class MatchboxLocalDuckDBLocalMixin:
 class MatchboxLocalDuckDB(
     MatchboxLocalDuckDBQueryMixin,
     MatchboxLocalDuckDBDataMixin,
-    MatchboxLocalDuckDBLocalMixin,
+    MatchboxLocalDuckDBCacheMixin,
     MatchboxLocalDBAdapter,
 ):
     """A DuckDB adapter for the local (client-side) Matchbox cluster store."""
@@ -727,8 +727,8 @@ class MatchboxLocalDuckDB(
         self._engine: Engine = db.create_local_engine(path)
         self._session_factory: sessionmaker = sessionmaker(bind=self._engine)
 
-        db.create_all_duckdb_safe(self._engine, orm.LOCAL_TABLES)
-        db.create_all_duckdb_safe(self._engine, orm.SHARED_TABLES)
+        db.create_db(self._engine, orm.LOCAL_TABLES)
+        db.create_db(self._engine, orm.SHARED_TABLES)
 
         self.all_clusters = _Countable(self._session_factory, orm.Clusters.count)
         self.source_clusters = _Countable(self._session_factory, _count_source_clusters)
--- a/src/matchbox/client/adapters/duckdb/adapter.py
+++ b/src/matchbox/client/adapters/duckdb/adapter.py
@@ -27,6 +27,7 @@ from sqlalchemy import (
 )
 from sqlalchemy.engine import Engine
 from sqlalchemy.orm import Session, sessionmaker
+from sqlalchemy.sql.selectable import Select
 
 from matchbox.client.adapters.duckdb import db, orm
 from matchbox.client.base import MatchboxLocalDBAdapter
@@ -117,20 +118,20 @@ def _materialise(
     session.execute(text(ddl))
 
 
-def _read_all(session: Session, physical_name: str) -> ArrowTable:
-    raw_conn = session.connection().connection.driver_connection
-    return raw_conn.sql(f'SELECT * FROM "{physical_name}"').to_arrow_table()
-
-
-def _read_filtered_by_keys(
-    session: Session, physical_name: str, keys: list[str]
-) -> ArrowTable:
-    """Read rows with a matching key, via SEMI JOIN pushdown."""
-    raw_conn = session.connection().connection.driver_connection
-    raw_conn.register("key_filter", pa.table({"key": keys}))
-    return raw_conn.sql(
-        f'SELECT t.* FROM "{physical_name}" t SEMI JOIN "key_filter" f ON t.key = f.key'
-    ).to_arrow_table()
+def _arrow_query(session: Session, stmt: Select) -> ArrowTable:
+    """Execute a read-only select via DuckDB's native Arrow export."""
+    connection = session.connection()
+    schema_translate_map = connection.get_execution_options().get(
+        "schema_translate_map"
+    )
+    compiled = stmt.compile(
+        dialect=connection.engine.dialect,
+        compile_kwargs={"literal_binds": True},
+        schema_translate_map=schema_translate_map,
+        render_schema_translate=True,
+    )
+    raw_conn = connection.connection.driver_connection
+    return raw_conn.sql(str(compiled)).to_arrow_table()
 
 
 def _dump_dynamic_table(session: Session, name: str) -> list[dict]:
@@ -247,16 +248,19 @@ class MatchboxLocalDuckDBQueryMixin:
             if limit is not None:
                 query_stmt = query_stmt.limit(limit)
 
-            rows = session.execute(query_stmt).all()
+            arrow_table = _arrow_query(session, query_stmt)
 
-        ids = [row.root_id for row in rows]
-        keys = [row.key for row in rows]
         if return_leaf_id:
-            leaf_ids = [row.leaf_id for row in rows]
-            return pa.table({"id": ids, "key": keys, "leaf_id": leaf_ids}).cast(
-                SCHEMA_QUERY_WITH_LEAVES
+            return (
+                arrow_table.select(["root_id", "key", "leaf_id"])
+                .rename_columns(["id", "key", "leaf_id"])
+                .cast(SCHEMA_QUERY_WITH_LEAVES)
             )
-        return pa.table({"id": ids, "key": keys}).cast(SCHEMA_QUERY)
+        return (
+            arrow_table.select(["root_id", "key"])
+            .rename_columns(["id", "key"])
+            .cast(SCHEMA_QUERY)
+        )
 
     def match(  # noqa: D102
         self,
@@ -477,23 +481,15 @@ class MatchboxLocalDuckDBDataMixin:
     def get_model_data(self, path: ModelStepPath) -> ArrowTable:  # noqa: D102
         with self._session() as session:
             step = orm.Steps.from_name(session, path.name, StepType.MODEL)
-            rows = session.execute(
+            arrow_table = _arrow_query(
+                session,
                 select(
                     tables.ModelEdges.c.left_id,
                     tables.ModelEdges.c.right_id,
                     tables.ModelEdges.c.score,
-                ).where(tables.ModelEdges.c.step_id == step.step_id)
-            ).all()
-
-        if not rows:
-            return SCHEMA_MODEL_EDGES.empty_table()
-        return pa.table(
-            {
-                "left_id": [r[0] for r in rows],
-                "right_id": [r[1] for r in rows],
-                "score": [r[2] for r in rows],
-            }
-        ).cast(SCHEMA_MODEL_EDGES)
+                ).where(tables.ModelEdges.c.step_id == step.step_id),
+            )
+        return arrow_table.cast(SCHEMA_MODEL_EDGES)
 
     def get_resolver_data(self, path: ResolverStepPath) -> ArrowTable:  # noqa: D102
         with self._session() as session:
@@ -505,13 +501,9 @@ class MatchboxLocalDuckDBDataMixin:
                 membership.c.root_id.label("parent_id"),
                 membership.c.leaf_id.label("child_id"),
             ).order_by(membership.c.root_id, membership.c.leaf_id)
-            rows = session.execute(stmt).all()
+            arrow_table = _arrow_query(session, stmt)
 
-        if not rows:
-            return SCHEMA_CLUSTERS.empty_table()
-        return pa.table(
-            {"parent_id": [r[0] for r in rows], "child_id": [r[1] for r in rows]}
-        ).cast(SCHEMA_CLUSTERS)
+        return arrow_table.cast(SCHEMA_CLUSTERS)
 
     def dump(self) -> MatchboxSnapshot:  # noqa: D102
         with self._session() as session:
@@ -599,9 +591,14 @@ class MatchboxLocalDuckDBCacheMixin:
                 raise MatchboxDataNotFound(table="raw_data", data=[path.name])
 
             physical_name = _raw_data_table_name(row.source_step_id)
+            raw_conn = session.connection().connection.driver_connection
             if keys is None:
-                return _read_all(session, physical_name)
-            return _read_filtered_by_keys(session, physical_name, keys)
+                return raw_conn.sql(f'SELECT * FROM "{physical_name}"').to_arrow_table()
+            raw_conn.register("key_filter", pa.table({"key": keys}))
+            return raw_conn.sql(
+                f'SELECT t.* FROM "{physical_name}" t '
+                'SEMI JOIN "key_filter" f ON t.key = f.key'
+            ).to_arrow_table()
 
     def cache_query(  # noqa: D102
         self, key: str, table: ArrowTable, depends_on: list[StepPath]
@@ -647,7 +644,9 @@ class MatchboxLocalDuckDBCacheMixin:
             ).scalar_one_or_none()
             if row is None:
                 return None
-            return _read_all(session, _query_cache_table_name(row.cache_id))
+            physical_name = _query_cache_table_name(row.cache_id)
+            raw_conn = session.connection().connection.driver_connection
+            return raw_conn.sql(f'SELECT * FROM "{physical_name}"').to_arrow_table()
 
     def drop_step_data(self, path: StepPath) -> None:  # noqa: D102
         with self._session() as session:
@@ -657,6 +656,8 @@ class MatchboxLocalDuckDBCacheMixin:
 
     def _cascade_invalidate(self, session: Session, step: "orm.Steps") -> None:
         """Drop a step's data and its descendants', and their query cache.
+
+        Responsibility for committing lies with the caller.
 
         Deletes are unconditional and type-agnostic: harmless no-ops for
         tables a step doesn't apply to. RawData and Clusters/Contains are
--- a/src/matchbox/client/adapters/duckdb/adapter.py
+++ b/src/matchbox/client/adapters/duckdb/adapter.py
@@ -1,0 +1,742 @@
+"""DuckDB implementation of the local (client-side) Matchbox cluster store.
+
+Shares SQL builders from matchbox.common.adapters.sql with the Postgres
+backend. Engine-specific code lives in db.py and orm.py.
+"""
+
+import base64
+from collections.abc import Callable
+from pathlib import Path
+
+import pyarrow as pa
+from pyarrow import Table as ArrowTable
+from sqlalchemy import (
+    BigInteger,
+    column,
+    delete,
+    exists,
+    func,
+    insert,
+    literal,
+    literal_column,
+    select,
+    text,
+)
+from sqlalchemy import (
+    table as sa_table,
+)
+from sqlalchemy.engine import Engine
+from sqlalchemy.orm import Session, sessionmaker
+
+from matchbox.client.adapters.duckdb import db, orm
+from matchbox.client.base import MatchboxLocalDBAdapter
+from matchbox.common.adapters.protocol import MatchboxLocalBackends, MatchboxSnapshot
+from matchbox.common.adapters.sql import tables
+from matchbox.common.adapters.sql.insert import (
+    select_cluster_map,
+    select_contains_pairs,
+    select_key_expansion,
+    select_new_cluster_hashes,
+    select_resolver_membership,
+)
+from matchbox.common.adapters.sql.query import (
+    assemble_matches,
+    build_matching_leaves_cte,
+    build_target_cluster_cte,
+    build_unified_query,
+    resolver_membership_subquery,
+)
+from matchbox.common.adapters.sql.resolver import (
+    build_expanded_leaves_subquery,
+    build_leaf_hash_groups_query,
+    hash_resolver_parents,
+)
+from matchbox.common.adapters.sql.snapshot import dump_tables, restore_tables
+from matchbox.common.arrow import (
+    SCHEMA_CLUSTERS,
+    SCHEMA_MODEL_EDGES,
+    SCHEMA_QUERY,
+    SCHEMA_QUERY_WITH_LEAVES,
+)
+from matchbox.common.dtos import (
+    Match,
+    ModelStepPath,
+    ResolverStepPath,
+    SourceStepPath,
+    Step,
+    StepPath,
+    StepType,
+)
+from matchbox.common.exceptions import (
+    MatchboxDataNotFound,
+    MatchboxDeletionNotConfirmed,
+)
+
+
+def _translate_lineage(
+    session: Session, lineage_names: list[str]
+) -> list[tuple[int, int | None]]:
+    """Translate step names from self.lineage() into (step_id, source_config_id).
+
+    Model steps are dropped: they have no source_config_id and aren't
+    queryable, unlike source and resolver steps.
+    """
+    translated: list[tuple[int, int | None]] = []
+    for name in lineage_names:
+        step = orm.Steps.from_name(session, name)
+        if step.type == StepType.MODEL.value:
+            continue
+        source_config_id = (
+            step.source_config.source_config_id if step.source_config else None
+        )
+        translated.append((step.step_id, source_config_id))
+    return translated
+
+
+_RAW_DATA_TABLE_PREFIX = "mb_raw_data__"
+_QUERY_CACHE_TABLE_PREFIX = "mb_query_cache__"
+
+
+def _raw_data_table_name(source_step_id: int) -> str:
+    """Physical table name for a source step's raw data."""
+    return f"{_RAW_DATA_TABLE_PREFIX}{source_step_id}"
+
+
+def _query_cache_table_name(cache_id: int) -> str:
+    """Physical table name for a cached query result."""
+    return f"{_QUERY_CACHE_TABLE_PREFIX}{cache_id}"
+
+
+def _materialise(
+    session: Session, physical_name: str, view_name: str, table: ArrowTable
+) -> None:
+    """Zero-copy materialise an ArrowTable as a physical DuckDB table."""
+    raw_conn = session.connection().connection.driver_connection
+    raw_conn.register(view_name, table)
+    ddl = f'CREATE OR REPLACE TABLE "{physical_name}" AS SELECT * FROM "{view_name}"'
+    session.execute(text(ddl))
+
+
+def _read_all(session: Session, physical_name: str) -> ArrowTable:
+    raw_conn = session.connection().connection.driver_connection
+    return raw_conn.sql(f'SELECT * FROM "{physical_name}"').to_arrow_table()
+
+
+def _read_filtered_by_keys(
+    session: Session, physical_name: str, keys: list[str]
+) -> ArrowTable:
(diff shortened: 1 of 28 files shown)


src/matchbox/common/factories/scenarios.py
```diff
@@ -1293,9 +1297,11 @@ def setup_scenario(
         # our new warehouse object
         dag_testkit = dag_testkit.model_copy(deep=True)
 
-        # Restore backend and write sources to warehouse
+        # Restore backend, write sources to warehouse, rebind if local
         backend.restore(snapshot=snapshot)
         _testkitdag_to_location(warehouse, dag_testkit)
+        if isinstance(backend, MatchboxLocalDBAdapter):
+            backend.bind(dag_testkit.dag)
```
leo-mazzone, line 1304: note to self: what does binding do?

src/matchbox/client/adapters/duckdb/adapter.py
```diff
@@ -0,0 +1,742 @@
...
+            rows = session.execute(query_stmt).all()
+
+        ids = [row.root_id for row in rows]
+        keys = [row.key for row in rows]
+        if return_leaf_id:
+            leaf_ids = [row.leaf_id for row in rows]
+            return pa.table({"id": ids, "key": keys, "leaf_id": leaf_ids}).cast(
+                SCHEMA_QUERY_WITH_LEAVES
+            )
+        return pa.table({"id": ids, "key": keys}).cast(SCHEMA_QUERY)
```
will-langdale: I think this can be tightened -- we can lose the Python round trip and go straight to Arrow.

test/fixtures/db.py
```diff
@@ -463,17 +465,29 @@ def empty_tracker() -> None:
...
+
+    An in-memory duckdb file is cheap enough to create fresh for every test.
+    """
+    return MatchboxLocalDuckDB()
+
+
+@pytest.fixture(scope="function")
+def backend_instance(
+    request: pytest.FixtureRequest, backend: str
+) -> MatchboxClusterStoreAdapter:
```
leo-mazzone, line 490: note to self: what does the hierarchy of classes now look like?

src/matchbox/client/adapters/duckdb/adapter.py
```diff
@@ -0,0 +1,742 @@
...
+            matched_rows=matched_rows,
+            source=source,
+            source_config_id=source_config_id,
+            targets=targets,
+            target_config_ids=target_config_ids,
+        )
+
+
+class MatchboxLocalDuckDBDataMixin:
+    """Data mixin for the local DuckDB adapter: protocol block, replace policy."""
```
will-langdale: I don't understand anything after the colon. Remove.

test/client/local/test_cache_mixin.py
```diff
@@ -0,0 +1,182 @@
...
+from matchbox.client.base import MatchboxLocalDBAdapter
+from matchbox.common.dtos import CollectionName, RunID, StepPath
+from matchbox.common.exceptions import MatchboxDataNotFound, MatchboxStepNotFoundError
+from matchbox.common.factories.scenarios import setup_scenario
+from test.fixtures.db import LOCAL_BACKENDS
+
+# Placeholder collection/run: local ignores both, only the step name matters.
+_NONEXISTENT_STEP = StepPath(
+    collection=CollectionName("local_test"), run=RunID(1), name="nonexistent"
+)
```
leo-mazzone, line 18: ignores them? What will the interface be for connecting to a DAG remotely?

test/client/local/test_cache_mixin.py
```diff
@@ -0,0 +1,182 @@
...
+@pytest.mark.parametrize("backend", LOCAL_BACKENDS)
+class TestRawData:
+    @pytest.fixture(autouse=True)
+    def setup(
+        self, backend_instance: MatchboxLocalDBAdapter, sqla_sqlite_warehouse: Engine
+    ) -> None:
+        self.backend: MatchboxLocalDBAdapter = backend_instance
+        self.scenario = partial(setup_scenario, warehouse=sqla_sqlite_warehouse)
+
+    def test_round_trip(self) -> None:
```
leo-mazzone, line 30: none of these tests have docstrings

test/client/local/test_cache_mixin.py
will-langdale: The `DAG` object owns the collection and run ID -- that information will come from there.

test/client/local/test_cache_mixin.py
leo-mazzone: tbf, all these tests are so small and clear

leo-mazzone: feel free to resolve

test/client/local/test_local_mixin.py
```diff
@@ -0,0 +1,182 @@
...
+            assert self.backend.get_model_data(model.path).num_rows > 0
+            assert self.backend.get_resolver_data(resolver.resolver.path).num_rows > 0
+
+            self.backend.drop_step_data(model.path)
+
+            assert self.backend.get_model_data(model.path).num_rows == 0
+            assert self.backend.get_resolver_data(resolver.resolver.path).num_rows == 0
+
+    def test_clears_dependent_cache_only(self) -> None:
+        """Only cache entries depending on the dropped step are cleared."""
```
leo-mazzone: not an English sentence

src/matchbox/client/adapters/duckdb/adapter.py
```diff
@@ -0,0 +1,742 @@
...
+                    select(candidate_contains.c.root, candidate_contains.c.leaf).where(
+                        ~exists(
+                            select(1).where(
+                                tables.Contains.c.root == candidate_contains.c.root,
+                                tables.Contains.c.leaf == candidate_contains.c.leaf,
+                            )
+                        )
+                    ),
+                )
+            )
```
will-langdale, line 477: Why does this follow a different order to PostgreSQL's `insert_clusters()`? Should something be factored out? Is `exists` going to work at scale?

src/matchbox/client/adapters/duckdb/adapter.py
```diff
@@ -0,0 +1,742 @@
...
+
+        if not rows:
+            return SCHEMA_MODEL_EDGES.empty_table()
+        return pa.table(
+            {
+                "left_id": [r[0] for r in rows],
+                "right_id": [r[1] for r in rows],
+                "score": [r[2] for r in rows],
+            }
+        ).cast(SCHEMA_MODEL_EDGES)
```
will-langdale: Can we avoid the Python round trip?

test/client/local/test_cache_mixin.py
```diff
@@ -0,0 +1,182 @@
...
+                "independent", pa.table({"a": [2]}), depends_on=[dh]
+            )
+
+            self.backend.drop_step_data(crn)
+
+            assert self.backend.get_cached_query("dependent") is None
+            assert self.backend.get_cached_query("independent") is not None
+
+    def test_raw_data_survives(self) -> None:
+        """drop_step_data doesn't touch RawData - it's canonical, not a cache."""
```
leo-mazzone, line 183: canonical?

src/matchbox/client/adapters/duckdb/adapter.py
```diff
@@ -0,0 +1,742 @@
...
+                membership.c.root_id.label("parent_id"),
+                membership.c.leaf_id.label("child_id"),
+            ).order_by(membership.c.root_id, membership.c.leaf_id)
+            rows = session.execute(stmt).all()
+
+        if not rows:
+            return SCHEMA_CLUSTERS.empty_table()
+        return pa.table(
+            {"parent_id": [r[0] for r in rows], "child_id": [r[1] for r in rows]}
+        ).cast(SCHEMA_CLUSTERS)
```
will-langdale: Python round trip.

test/client/local/test_local_mixin.py
```diff
@@ -0,0 +1,182 @@
+"""Tests for MatchboxLocalDuckDBLocalMixin: raw data, query cache, cascade."""
```
leo-mazzone: local/test_local_mixin.py

vs

local/data_mixin.py

?

leo-mazzone: Perhaps MatchboxLocalDuckDBLocalMixin is misnamed

src/matchbox/client/adapters/duckdb/adapter.py
```diff
@@ -0,0 +1,742 @@
...
+            physical_names = [
+                _raw_data_table_name(row["source_step_id"]) for row in data["raw_data"]
+            ] + [
+                _query_cache_table_name(row["cache_id"]) for row in data["query_cache"]
+            ]
+            # dump_tables only covers catalog rows - physical tables need
+            # dumping separately, same row format.
+            data["__dynamic_cache_tables__"] = {
+                name: _dump_dynamic_table(session, name) for name in physical_names
+            }
```
will-langdale, line 520: Could we solve this by just adding physical tables to the catalogue, and removing if steps change? Why handle separately?

src/matchbox/client/adapters/duckdb/adapter.py
```diff
@@ -0,0 +1,742 @@
...
+
+        self.clear(certain=True)
+
+        with self._session() as session:
+            restore_tables(
+                session, [*orm.LOCAL_TABLES, *orm.SHARED_TABLES], snapshot.data
+            )
+            for name, rows in snapshot.data.get("__dynamic_cache_tables__", {}).items():
+                _restore_dynamic_table(session, name, rows)
+            session.commit()
```
will-langdale, line 537: Same here. Can we simplify and remove these helpers by using the catalogue?

src/matchbox/client/adapters/duckdb/adapter.py
```diff
@@ -0,0 +1,742 @@
...
+    def drop(self, certain: bool) -> None:  # noqa: D102
+        if not certain:
+            raise MatchboxDeletionNotConfirmed(
+                "This operation will drop the entire database and recreate it. "
+                "Rerun with certain=True to continue."
+            )
+        db.drop_dynamic_cache_tables(
+            self._engine, (_RAW_DATA_TABLE_PREFIX, _QUERY_CACHE_TABLE_PREFIX)
+        )
+        db.drop_all_duckdb(self._engine, [*orm.SHARED_TABLES, *orm.LOCAL_TABLES])
```
will-langdale: Used once. Inline.

src/matchbox/client/adapters/duckdb/db.py
```diff
@@ -0,0 +1,81 @@
...
+]
+
+
+def create_local_engine(path: Path | None) -> Engine:
+    """Create a duckdb engine, in-memory and ephemeral if no path is given."""
+    url = f"duckdb:///{path}" if path is not None else "duckdb:///:memory:"
+    return create_engine(url).execution_options(schema_translate_map={"mb": None})
+
+
+def create_all_duckdb_safe(engine: Engine, order: list[Table]) -> None:
```
will-langdale: We're in a duckdb module. Rename to `create_db()`.

test/client/local/test_data_mixin.py
```diff
@@ -0,0 +1,76 @@
+"""Tests for MatchboxLocalDuckDBDataMixin: replace-on-rerun.
+
+Local stores replace-on-rerun, the deliberate inversion of the server's
```
leo-mazzone, line 3: doesn't the server also replace on resync? I guess only if the signature or whatever is called changes?

src/matchbox/client/adapters/duckdb/db.py
```diff
@@ -0,0 +1,81 @@
...
+
+        for tbl, col_name in SEQUENCE_COLUMNS:
+            if tbl in order:
+                conn.exec_driver_sql(
+                    f'ALTER TABLE "{tbl.name}" ALTER COLUMN "{col_name}" '
+                    f"SET DEFAULT nextval('{tbl.name}_{col_name}_seq')"
+                )
+
+
+def drop_all_duckdb(engine: Engine, tables: list[Table]) -> None:
```
will-langdale: If we keep, rename to `drop_db()`. Given it's not even used in `drop_dynamic_cache_tables()`, I don't think it's worth the code.

test/client/local/test_data_mixin.py
leo-mazzone: I think we really need an idea of where we're going with regards to the server

src/matchbox/client/adapters/duckdb/adapter.py
```diff
@@ -0,0 +1,742 @@
...
+            self._cascade_invalidate(session, step)
+            session.commit()
+
+    def _cascade_invalidate(self, session: Session, step: "orm.Steps") -> None:
+        """Drop a step's data and its descendants', and their query cache.
+
+        Deletes are unconditional and type-agnostic: harmless no-ops for
+        tables a step doesn't apply to. RawData and Clusters/Contains are
+        left alone - canonical state and content-addressed rows, not
+        caches. Query cache invalidation is selective, via QueryCacheStep.
```
will-langdale, line 665: Worth noting the caller has to commit in the docstring.

test/client/local/test_data_mixin.py
```diff
@@ -0,0 +1,76 @@
...
+            assert before > 0
+
+            self.backend.insert_resolver_data(
+                path=resolver.resolver.path, data=resolver.resolver.results.to_arrow()
+            )
+
+            assert (
+                self.backend.get_resolver_data(resolver.resolver.path).num_rows
+                == before
+            )
```
leo-mazzone, line 76: cascading isn't checked here?

src/matchbox/client/base.py
will-langdale, line 1: Drop backticks from docstrings in this file.

src/matchbox/client/base.py
```diff
@@ -0,0 +1,205 @@
...
+
+        allowed = set(sources)
+        for source in sources:
+            allowed.update(_ancestor_depths(children, source))
+        ancestors = {
+            name: depth for name, depth in ancestors.items() if name in allowed
+        }
+
+    ordered = sorted(ancestors.items(), key=lambda pair: (pair[1], pair[0]))
+    return [resolver, *(name for name, _ in ordered)]
```
will-langdale, line 71: Are we duplicating `DAG._topological_sort()` and the method that rely on it? How can we reduce this repetition? 

Ideas:

* `bind()` uses a whole `DAG`, and so can defer to its methods
* An abstract `common.graph` module that centralises these algorithms and any coverage of them

src/matchbox/client/base.py
```diff
@@ -0,0 +1,205 @@
...
+            children.setdefault(parent, []).append(name)
+    return list(_ancestor_depths(children, step))
+
+
+class MatchboxLocalSettings(BaseModel):
+    """Settings for a local (client-side) Matchbox backend."""
+
+    backend_type: MatchboxLocalBackends = MatchboxLocalBackends.DUCKDB
+    path: Path | None = None
+    """Path to the local database file. None means in-memory and ephemeral."""
```
will-langdale: This is pydantic, use `Field(description=)`.

src/matchbox/client/base.py
```diff
@@ -0,0 +1,205 @@
...
+    def bind(self, graph_source: HasGraph) -> None:
+        """Bind a live source of step topology.
+
+        Stored by reference, not copied: lineage()/descendants() read
+        graph_source.graph fresh on every call, so later mutations are
+        picked up automatically.
+        """
+        self._graph_source = graph_source
+
+    def _graph(self) -> dict[StepName, list[StepName]]:
```
will-langdale, line 123: Is this a `@property`?

src/matchbox/common/adapters/sql/tables.py
```diff
@@ -108,7 +117,13 @@
 ModelEdges = Table(
     "model_edges",
     METADATA,
-    Column("result_id", BigInteger, primary_key=True, autoincrement=True),
+    Column(
+        "result_id",
+        BigInteger,
+        primary_key=True,
+        autoincrement=False,
```
leo-mazzone, line 124: concerned about this autoincrement=False

src/matchbox/client/adapters/duckdb/db.py
```diff
@@ -0,0 +1,81 @@
...
+    with engine.begin() as conn:
+        for tbl, col_name in SEQUENCE_COLUMNS:
+            if tbl in order:
+                conn.exec_driver_sql(
+                    f'CREATE SEQUENCE IF NOT EXISTS "{tbl.name}_{col_name}_seq"'
+                )
+
+        for tbl in order:
+            ddl = str(
+                CreateTable(tbl, include_foreign_key_constraints=[]).compile(
```
joshwong-cs, line 49: CreateTable has an `if_not_exists` arg, do we want to toggle it on here?

test/client/adapters/duckdb/test_duckdb_core.py
```diff
@@ -0,0 +1,57 @@
...
+
+        with matchbox_local_duckdb._engine.connect() as conn:
+            names = {
+                row[0]
+                for row in conn.exec_driver_sql(
+                    "SELECT table_name FROM duckdb_tables()"
+                ).fetchall()
+            }
+        prefixes = ("mb_raw_data__", "mb_query_cache__")
+        assert not any(n.startswith(prefixes) for n in names)
```
leo-mazzone, line 33: why is this so low level?

test/client/adapters/duckdb/test_duckdb_core.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def test_dump_restore_round_trips_dynamic_tables(
+    matchbox_local_duckdb: MatchboxLocalDuckDB, sqla_sqlite_warehouse: Engine
+) -> None:
+    """dump()/restore() carry RawData/QueryCache row content, not just pointers."""
```
leo-mazzone, line 39: pointers?

src/matchbox/server/postgresql/utils/insert.py
```diff
@@ -215,17 +206,16 @@ def insert_model_edges(
...
                 insert(ModelEdges)
                 .from_select(
                     ["step_id", "left_id", "right_id", "score"],
-                    edges_select,
+                    select(
+                        literal(step.step_id, BigInteger).label("step_id"),
+                        incoming_edges.c.left_id,
+                        incoming_edges.c.right_id,
+                        incoming_edges.c.score,
+                    ),
```
will-langdale, line 218: Revert to a separate var -- no need to change, was clear before.

test/client/adapters/duckdb/test_duckdb_core.py
will-langdale, line 1: If the catalogue suggestion works, I think we could be able to drop these tests.

test/client/local/test_data_mixin.py
will-langdale: Once you write a run, that's it, isn't it? Or you can overwrite it for a while then once it's default you need a new run.

src/matchbox/client/adapters/duckdb/adapter.py
```diff
@@ -0,0 +1,742 @@
...
+            )
+        db.drop_dynamic_cache_tables(
+            self._engine, (_RAW_DATA_TABLE_PREFIX, _QUERY_CACHE_TABLE_PREFIX)
+        )
+        db.drop_all_duckdb(self._engine, [*orm.SHARED_TABLES, *orm.LOCAL_TABLES])
+        db.create_all_duckdb_safe(self._engine, orm.LOCAL_TABLES)
+        db.create_all_duckdb_safe(self._engine, orm.SHARED_TABLES)
+
+
+class MatchboxLocalDuckDBLocalMixin:
```
will-langdale: As @leo-mazzone notes elsewhere, this is an awful name.

Let's rename the ABC `MatchboxCacheMixin`, this can be `MatchboxLocalDuckDBCacheMixin`, and the test file can be `local/test_cache_mixin.py`.

test/client/local/test_local_mixin.py
will-langdale: Agree, will go for `MatchboxLocalDuckDBCacheMixin`.

test/client/local/test_cache_mixin.py
will-langdale: I like docstrings and will add.

test/client/local/test_data_mixin.py
will-langdale: `TestDropStepData.test_cascades` covers at a higher level.

test/client/local/test_local_mixin.py
will-langdale: "Only the caches of descendents of the dropped step are cleared"? In the test name too.

test/client/local/test_cache_mixin.py
will-langdale: I actually don't like this behaviour. We might add or remove fields, want to run it again with fresh data. I think it should drop with the step.

src/matchbox/client/adapters/duckdb/adapter.py
```diff
@@ -0,0 +1,742 @@
...
+from matchbox.common.exceptions import (
+    MatchboxDataNotFound,
+    MatchboxDeletionNotConfirmed,
+)
+
+
+def _translate_lineage(
+    session: Session, lineage_names: list[str]
+) -> list[tuple[int, int | None]]:
+    """Translate step names from self.lineage() into (step_id, source_config_id).
```
leo-mazzone, line 80: self.lineage()? This ain't a class

src/matchbox/client/adapters/duckdb/adapter.py
```diff
@@ -0,0 +1,742 @@
...
+    def restore(self, snapshot: MatchboxSnapshot) -> None:  # noqa: D102
+        if snapshot.backend_type != MatchboxLocalBackends.DUCKDB:
+            raise TypeError(
+                f"Cannot restore {snapshot.backend_type} snapshot to duckdb backend"
+            )
+
+        self.clear(certain=True)
+
+        with self._session() as session:
+            restore_tables(
```
joshwong-cs, line 532: need to reset each sequence in SEQUENCE_COLUMNS to MAX(pk) + 1 since restore_tables() inserts explicit primary keys

src/matchbox/common/adapters/sql/insert.py
```diff
@@ -0,0 +1,74 @@
+"""Ingest select builders, shared across relational backends.
```
leo-mazzone, line 1: more information needs to be given about

* the structure of the incoming `TableClause` and the other two `Subquery`s
* conceptually, what are these functions doing

leo-mazzone: otherwise it's really difficult to reuse functions from this file. Does this help my use case? Who knows

src/matchbox/common/adapters/sql/insert.py
```diff
@@ -0,0 +1,74 @@
...
+
```
leo-mazzone, line 2: these functions assume a specific ORM structure, which I guess is fair for SQL adapters. But it seems we're also increasingly assuming all adapters will be relational?

src/matchbox/common/adapters/sql/snapshot.py
```diff
@@ -0,0 +1,86 @@
+"""Generic table dump/restore, shared across relational backends.
+
+Handles only data movement: selecting every row and base64-encoding bytes
+columns for JSON-safety, or decoding and bulk-inserting them back.
+Engine-specific concerns stay with the caller.
```
leo-mazzone, line 5: for example? What's outside "data movement"? ID sequences?

src/matchbox/client/adapters/duckdb/db.py
will-langdale: Renamed. Not removing yet just in case the catalogue work does.

src/matchbox/server/postgresql/utils/insert.py
will-langdale: Disagree now: matches in both adapters.

src/matchbox/client/base.py
```diff
@@ -0,0 +1,209 @@
...
+    ) -> Table:
+        """Get raw warehouse rows for a source step, optionally filtered by key."""
+        ...
+
+    @abstractmethod
+    def cache_query(self, key: str, table: Table, depends_on: list[StepPath]) -> None:
+        """Cache cleaned query data, keyed by query config plus upstream fingerprints.
+
+        Disposable: unlike insert_raw_data, this is just a cache, and can be
+        recomputed at any time.
```
leo-mazzone, line 167: what does "at any time" mean?

src/matchbox/client/base.py
```diff
@@ -0,0 +1,209 @@
...
+
+class MatchboxLocalDBAdapter(MatchboxClusterStoreAdapter, ABC):
+    """An abstract base class for Matchbox local (client-side) database adapters.
+
+    Extends MatchboxClusterStoreAdapter (the query block, data block, and
+    cluster counts) with the local-only surface: raw warehouse data, a
+    disposable query cache, and explicit cascade invalidation.
+
+    Local stores replace-on-rerun (with cascade invalidation of descendant
+    steps), unlike server backends, which are write-once per step per run.
```
leo-mazzone, line 105: like pointed out by many other comments elsewhere, this really does not clarify the invalidation model. I am confusion

src/matchbox/client/base.py
```diff
@@ -0,0 +1,209 @@
...
+
+    The adapter is bound to a live DAG via bind(). lineage() and
+    descendants() read its graph fresh on every call, so steps added
+    later are picked up automatically.
+    """
+
+    _graph_source: HasGraph | None = None
+
+    def bind(self, graph_source: HasGraph) -> None:
```
leo-mazzone, line 114: `bind_graph`?

src/matchbox/client/adapters/duckdb/adapter.py
will-langdale: Might still be needed depending on the catalogue work.

src/matchbox/client/adapters/duckdb/orm.py
```diff
@@ -0,0 +1,228 @@
...
+    """
+
+    @classmethod
+    def count(cls, session: Session) -> int:
+        """Counts the number of rows in the table."""
+        return session.execute(select(func.count()).select_from(cls)).scalar_one()
+
+
+class Steps(LocalBase):
+    """Local step registry. No run_id/collection/upload_stage: one file is one run."""
```
leo-mazzone, line 63: Get rid of

> No run_id/collection/upload_stage

src/matchbox/client/adapters/duckdb/orm.py
```diff
@@ -0,0 +1,228 @@
...
+            raise MatchboxStepTypeError(
+                step_name=name,
+                step_type=StepType(step.type),
+                expected_step_types=[expected_type],
+            )
+        return step
+
+
+class SourceConfigs(LocalBase):
+    """Degenerate one-row-per-source table.
```
leo-mazzone, line 109: degenerate?

src/matchbox/client/adapters/duckdb/orm.py
```diff
@@ -0,0 +1,228 @@
...
+
+    Keeps the join shape identical to the server's, without needing the
+    rest of a real SourceConfig.
```
leo-mazzone, line 112: I don't understand what this means. Are these IDs the same you'll find on server?

src/matchbox/client/adapters/duckdb/orm.py
```diff
@@ -0,0 +1,228 @@
...
+    )
+    step_id: Mapped[int] = mapped_column(
+        BigInteger, ForeignKey("steps.step_id"), primary_key=True
+    )
+
+
+class Clusters(CountMixin, LocalBase):
+    """Table of indexed data and clusters that match it."""
+
+    __table__ = tables.Clusters.to_metadata(LOCAL_METADATA, schema=None)
```
leo-mazzone, line 188: is schema=None right? Where does the schema come from for these tables?

src/matchbox/client/adapters/duckdb/orm.py
```diff
@@ -0,0 +1,228 @@
...
+    QueryCache.__table__,
+    QueryCacheStep.__table__,
+]
+SHARED_TABLES = [
+    Clusters.__table__,
+    ClusterSourceKey.__table__,
+    Contains.__table__,
+    ModelEdges.__table__,
+    ResolverClusters.__table__,
+]
```
leo-mazzone, line 228: what are these two lists for?

src/matchbox/client/base.py
will-langdale: Discussed with @leo-mazzone.

* Backend belongs to the DAG
* Like a `Step`, even though it belongs to the DAG, the backend can query the DAG
* So topological algorithms belong to the DAG
* Note: ensuring that DAG and backend are always one is crucial. If you delete a backend from a DAG, that backend shouldn't be able to call the DAG

src/matchbox/client/adapters/duckdb/adapter.py
will-langdale: No. We can reflect to dump, but not to restore, where we infer the schema from the JSON because the tables aren't defined. You need to create the table with the right columns... which is exactly what restore does, with the materialise helper.

src/matchbox/client/adapters/duckdb/adapter.py
will-langdale: It isn't, but I like the symmetry. Keeping it.

src/matchbox/client/adapters/duckdb/adapter.py
will-langdale: Removed.

src/matchbox/client/adapters/duckdb/adapter.py
will-langdale: Yes. Removed.

src/matchbox/client/adapters/duckdb/adapter.py
will-langdale: Removed.

src/matchbox/client/adapters/duckdb/adapter.py
will-langdale: Noted.

